PR #5, opened 2016-01-06
PCHR-572: Remove initial join and termination dates from the filter i…

Agares4:
…n taskassignments/key-dates

Before:
screenshot 27
After:
screenshot 26


AkA84:
Approved

PR #11, opened 2016-01-13
PCHR-554: show all keydates for next 7 days in reminder email

nishant-bhorodia:
Issue: 
Keydates sections was not displaying all keydates
t a-keydates-before

Solution:
Fetch all keydates irrespective of contactID
t a-keydates-after


Agares4:
Approved

PR #12, opened 2016-01-13
PCHR-635: Previous assignee is informed when task's assignee is changed.

Agares4:
An email notification was not sent to the previous assignee, when the assignee of a task was changed. This PR adds that notification.


coldrunKacper:
Approved.

PR #13, opened 2016-01-13
PCHR-550: Notify  assignee and creator on task deletion

nishant-bhorodia:
No notification was generated on task deletion. This PR generates notification email for Creator and Assignee when task is deleted.

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/api/v3/Task.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/api/v3/Task.php
@@ -322,6 +322,7 @@ function civicrm_api3_task_delete($params) {
   }
   
   if (CRM_Activity_BAO_Activity::deleteActivity($params)) {
+    CRM_Tasksassignments_Reminder::sendReminder($params['id']);
     return civicrm_api3_create_success(1, $params, 'activity', 'delete');
   }
   else {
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.tasksassignments/api/v3/Task.php
```diff
@@ -322,6 +322,7 @@ function civicrm_api3_task_delete($params) {
   }
   
   if (CRM_Activity_BAO_Activity::deleteActivity($params)) {
+    CRM_Tasksassignments_Reminder::sendReminder($params['id']);
```
Agares4: shouldn't that be done before the activity is deleted?

nishant-bhorodia: Thanks @Agares4 
If we did that and activity is not deleted because of some error, wouln't that cause issues maybe ?

Agares4: Yes, it might cause some issues. But the sendReminder method needs to find an activity by its ID, so are you sure it even works this way? :O

nishant-bhorodia:
Thank you @Agares4 , you are right !!

Moved it before activity deletion step.

Agares4:
Approved

PR #14, opened 2016-01-14
PCHR-778 Task Modal Design & Save And New feature

kkalamarski:
Invision design: https://projects.invisionapp.com/d/main#/console/5036643/114143683/comments

Design Before:
778-before

Design After:
778-after

Since we don't have a dedicated directive for contact lookup, I disabled caching for contact fields:
lookup-no-cache

Also in this PR is implemented functionality for new button 'Save And New':
save-and-new

CSS:
I needed to override ui-select directive's design to make it fit the design from [invision](https://projects.invisionapp.com/d/main#/console/5036643/114143683/comments).
Unfortunately needed to use few !important, but it's to be fixed when we have a new directive.


uk.co.compucorp.civicrm.tasksassignments/js/services/contact.js
```diff
@@ -1,101 +1,115 @@
...
+                        if (UtilsService.errorHandler(data, 'Unable to fetch contact list', deferred)) {
+                            return
+                        }
 
-                    if (UtilsService.errorHandler(data,'Unable to fetch contact list',deferred)) {
-                        return
-                    }
+                        var contacts = [];
+                        
+                        for (var i in data.values) {
```
AkA84: Not an issue, but I think this could be more tersely rewritten as
```js
deferred.resolve(data.values.map(function (contact) {
    return {
        label: contact.display_name,
        description: [contact.email],
        id: contact.contact_id,
        icon_class: contact.contact_type
    };
}));

File: uk.co.compucorp.civicrm.tasksassignments/js/services/contact.js [hunk 2]
```diff
@@ -1,101 +1,115 @@
...
+                            contacts.push({
+                                label: data.values[i].display_name,
+                                description: [data.values[i].email],
```
AkA84: Is this correct? `description` is a single-item array containing an `email` value?

kkalamarski: That's how it was before I started working on that. I don't have time to rebuild all structure, that's why I decided to leave it for later.

AkA84:
Approved

uk.co.compucorp.civicrm.tasksassignments/js/services/contact.js
kkalamarski: In this case data.values is not an array.

AkA84: Oh right, I think the API call is missing the `sequential` param then, which I think makes sure the response is in an array format. Nevermind then!

PR #19, opened 2016-01-26
PCHR-434: Changed behaviour of my-documents and my-tasks for not-logg…

Agares4:
…ed users

civicrm/tasksassignments/my-tasks and civicrm/tasksassignments/my-documents now redirect to login in case of users who are not logged in


coldrunKacper:
Approved

PR #23, opened 2016-02-04
PCHR-839: Setup test environment

AkA84:
This PR sets up a test environment for the Tasks and Assignments app

#### Things done
* Fixed line endings on some of the templates. They were set to "Mac OS" instead of "Unix", and that resulted an error when trying to convert the html files into js via the `ng-html2js` preprocessor
* Added `test-main.js`, the entry file for the tests (bootstraps the whole test suite via requireJS)
* Added a `global.js` files, which is used to define the global variables (like `CRM`) that the tests should have access to
* Added an example test


kkalamarski:
Approved

PR #24, opened 2016-02-05
PCHR-840: Upgrade to AngularJS v1.4.9

AkA84:
*(Related to https://github.com/civicrm/civihr/pull/848)*

The newer angular version doesn't like filters with names using the dot.notation, so changing the names to camelCase solves the issue


kkalamarski:
Approved

PR #26, opened 2016-02-12
PCHR-825: Removed 'get started' text from reminder email footer

nishant-bhorodia:
Removed 'get started' text from reminder email footer


AkA84:
Approved

PR #28, opened 2016-02-24
PCHR-634: Due date modified

kkalamarski:
## Problem:
Dates were sent to backend with timezone information, which caused bug we've already fixed in HRJobRoles. (Date was going back by 1 day)

If wrong date was saved to database, also notification email had wrong date.

## Solution:
Using `moment()` to parse date to format `YYYY-MM-DD`, which is correctly read by backend.
Now date is correctly saved to database, so date in email notification is also correct.

Also removed CommonJS version of `angular-date` and implemented `common.angularDate` (AMD).


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/modules/run.js
```diff
@@ -9,16 +9,22 @@ define([
     'tasks-assignments/vendor/text-angular',
     'tasks-assignments/modules/config',
     'tasks-assignments/modules/settings',
+    // TODO use only one router
```
AkA84: Why is the app using two different routers?

kkalamarski: I have no idea. It was like this when I started working on that. But I'll refactor it when I work on certain modules.

AkA84: Would it be a big refactoring? Could it be done as part of this PR?

kkalamarski: Actually it can be done.

AkA84:
Approved

PR #30, opened 2016-03-30
PCHR-478: Copy task assignee

kkalamarski:
## Problems:

**1. Couldn't save form with disabled fields.**
They were required to be filled up, but it was not possible.

**2. Copy Button**
2.1 - Documents - Assigned to is copying assignee from tasks, instead of the first row of document
2.2 - Dates copy are not working for both tasks and documents 
2.3 - Copy button for the assignee works for the first time, but if you change the value again and copy, nothing happened
2.4 - Nice to have: If we could show hover over tool tip. See example from default Civi.

**3. Select list should have square edge.** 
Date picker seemed too wide for both Key date and Due Dates.

## Solutions
**1. Fixed `ng-require` directive to correctly validate required fields.**
1fixed

**2. Copy button now works correctly.**
copy2

2.4. Added tooltip.
2 4 fixed

**3. Using  `<ui-select>` instead of HTML5 `<select>`.**
square


AkA84:
You are adding new functionality to `ModalAssignmentCtrl`, you need also to write tests for that before the PR can be approved

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-assignment_test.js
```diff
@@ -0,0 +1,104 @@
...
+                ContactService: ContactService,
+                data: {},
+                config: {},
+                settings: {},
+                HR_settings: HR_settings
+            });
+        }));
+
+        it('Controller should be defined', function () {
+            expect(ctrl).toBeDefined()
```
AkA84: I'm not sure this is a meaningful test, if the `ModalAssignmentCtrl` would not exist, then the entire test suite would throw an error, i don't believe you will get into this test example anyway

If you want to test this, maybe you could rewrite it as something like this:
```js
expect(initController).not.toThrow();

function initController() {
    ctrl = $controller('ModalAssignmentCtrl', { /* .. */ });
}
```

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-assignment_test.js
```diff
@@ -0,0 +1,104 @@
...
+        });
+
+        describe('Bulk Assign', function(){
+            beforeEach(function(){
+                angular.extend(scope, Mock);
+            });
+
+            it('Test copyAssignee method', function(){
```
AkA84: Please write tests in a more readable, english-like way. For example:
```js
describe('copyAssignee()', function () {
    var list, assigneeId = [2];

    beforeEach(function () {
        list = scope.taskList;
        list[0].assignee_contact_id = assigneeId;

        scope.copyAssignee(list);
    });

    it("assigns the assignee's id of the first item to the whole list", function () {
        expect(list.every(function (item) {
            return item.assignee_contact_id.toEqual(assigneeId);
        })).toBe(true);
    });
});
```
_(I don't think there is the necessity here to test the "before" status, it's pretty unlikely that all the items have already the same assignee id, and in any case you are the one who wrote the mocked list, and thus the initial status of it)_

The goal here is to have a test suite for components that you can quickly read through (by just focusing on the `describe` and `it` description) and get a good idea of what they do, what their interface is, etc. A simple `Test copyAssignee method` doesn't really give any information of any kind to the reader

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-assignment_test.js
```diff
@@ -0,0 +1,104 @@
...
+ $log: $log,
+ $filter: $filter,
+ AssignmentService: AssignmentService,
+ TaskService: TaskService,
+ DocumentService: DocumentService,
+ ContactService: ContactService,
+ data: {},
+ config: {},
+ settings: {},
+ HR_settings: HR_settings
```
AkA84: I'm not sure what these lines are for?

kkalamarski: This is outdated. Not sure why it showed to you. It's removed now.

AkA84:
Approved

PR #31, opened 2016-04-01
PCHR-879: Issue with ui-select with AngularJS v1.5

AkA84:
## Problem
issue

The ui.select library is throwing errors now that we have moved to AngularJS v.15 ([PCHR-870](https://github.com/civicrm/civihr/pull/871)). This seems kind of a big (recurrent?) issue for the library which currently still have to be addressed

## Solution
There is a (seemingly working) workaround to this, which consists in applying to the directive tags same-named classes:

```html
<ui-select-match class="ui-select-match"...>
<ui-select-choices class="ui-select-choices" ...>
```

for some reason this works, and so I have gone through every place where the directive is used and implement those classes manually


kkalamarski:
Approved

PR #33, opened 2016-04-04
PCHR-853: Fix overlapping edit in place

AkA84:
It wasn't possible to click "Follow Link" if it was place above another edit-in-place element. A simple fix on the `z-index` property solves the problem

##### Before
before

##### After
after

Also there is a fix related to #31, where the class-based fix is implemented on one of our custom directive


davialexandre:
@AkA84 Approved

PR #34, opened 2016-04-05
PCHR-632: Fix spelling

AkA84:
In the documents and tasks forms, "Assignment" was mispelled

##### Before
before-tasks
before-documents

##### After
after-documents
after-tasks


omar-compucorp:
Approved

PR #35, opened 2016-04-05
PCHR-637: Add more information to the Assignment lookup select dropdown

AkA84:
Adding more informations (id, status, creation date, subject) to the Assignment lookup select dropdown, to make it work similar to the "File to Case" feature in default CiviCRM

#### Before
before

#### After
task-modal


omar-compucorp:
Approved

PR #36, opened 2016-04-05
PCHR-863: Show all task data in Edit Task modal

AkA84:
#### Problem
When the "Edit Task" modal was open, most of the data of the task wasn't displayed on the modal

before

#### Solution
The issue is solved in two steps:
1. The `ng-show` logic wasn't implemented correctly, that is the fields weren't shown if the data was already there
2. For the lookup lists loaded via ajax (assignments, assignees, target), the list given to the lookup directive must come pre-loaded with the item already selected for the task

Additionally, the layout of the static text for the task name has been fixed, as originally it looked a bit too squashed between the columns

after


uk.co.compucorp.civicrm.tasksassignments/views/modal/task.html
```diff
@@ -74,14 +75,14 @@
             </div>
             <!-- Subject -->
             <div class="row">
-                <div class="col-xs-12 col-sm-6" ng-show="!showFieldSubject">
+                <div class="col-xs-12 col-sm-6" ng-show="!task.subject && !showFieldSubject">
                     <span class="form-control-static">
                         <a href class="btn btn-indent" ng-click="showFieldSubject = true;">
```
omar-compucorp: i don't know but i think it is better for href to have a value .. something like href="javascript:void(0);"

AkA84: That's AngularJS preferred syntax, you don't give `href` a value and it knows that it's a link that shouldn't do anything if clicked

omar-compucorp:
while i can't fully review it since i am not that familier with angular , but i think it looks good but don't forget to squash the commits .

:)

danilo-compucorp:
Looks good. PR approved.

PR #37, opened 2016-04-07
PCHR-853: Fix z-index for edit in place buttons

AkA84:
## Problem
The edit in place buttons would not appear above everything else if another edit in place element was placed just right below.

before

## Solution
Applying a higher z-index value solves the problem

after


kkalamarski:
Approved

PR #38, opened 2016-04-07
PCHR-778: Add task design fix

kkalamarski:
## 1. Problem
Modal was too wide.
width_before

### Solution
Modal width was changed from `600px` to `500px` to meet the design.
width_after

## 2. Problem
Due date should be set to today's date by default (pre-filled).
date_before

### Solution
Due date is now set to today's date.
date_after

## 3. Problem
When a task is created, Target and Assignee fields are empty in the dashboard, unless you refresh the page.

**Reason:**
method `updateCache()` was altering cache with object that was lacking `display_name` field. Therefore template was trying to display field that no longer exist in model. 
missing_name_before

### Solution
`display_name` field was added to updateCache method. 
missing_name_after


kkalamarski:
Should be merged to develop.

PR #39, opened 2016-04-07
PCHR-862: Random assignee is shown in new task notification email

AkA84:
A random assignee was appearing in the assignee list of the email that is sent when a task is created.

## Problem
This additional assignee was coming from the `$previousAssigneeId` whose value was obtained by calling the 'ActivityContact' api filtered by the current task id. The problem was that the call was being made  even if the task was created, not edited (that is, when the task didn't have an id with which we can filter the Activities). As a result the api would return **all** the activities, and the `$previousAssigneeId` value would end up being just the `contact_id` of the first activity whose `record_type_id` value was 1

## Solution
The api call is now made only if the task has an id. If it has, then it makes sense to look for the previous assignee; if it doesn't, then it means that the task has just been created so the previous assignee is just `null`


davialexandre:
@AkA84 Approved

PR #40, opened 2016-04-08
PCHR-635: Original assignee is not distinct from the new assignee

AkA84:
## Problem
When a task is assigned from a person to another, the old assignee is listed in the "Assignee" list alongside the new one, making harder to tell which one is which

before

## Solution
The previous assignee is now displayed separately from the new one

after

### Code refactoring
Also the `sendReminder` static method has been refactored (the focus was kept still on the scope of the ticket, so no parts of the code unrelated to the issue were changed) into smaller methods.

One noteworthy change was how the activity contact's emails were fetched. Before it was done with two api calls:
```php
$contactResult = civicrm_api3('Contact', 'get', array(
  'return' => 'display_name',
  'id' => array('IN' => $value['ids']),
));

$emailResult = civicrm_api3('Email', 'get', array(
  'return' => 'contact_id,email',
  'contact_id' => array('IN' => $value['ids']),
  'is_primary' => 1,
));
```
After the refactoring, the same result is achieved with one api call only, by chaining the two calls together:
```php
$contacts = civicrm_api3('Contact', 'get', array(
    'return' => 'display_name',
    'id' => array('IN' => $contactIds),
    'api.Email.getsingle' => array(
        'contact_id' => "\$value.contact_id",
        'is_primary' => true,
        'return' => array('contact_id', 'email')
    )
))['values'];
```

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
@@ -36,19 +36,75 @@ class CRM_Tasksassignments_Reminder
             self::$_relatedExtensions['appraisals'] = true;
         }
     }
-    
+
+    /**
+     * Gets the contacts starting with the given ids
+     *
+     * It also chains a call to the 'Email' api and puts the value in
+     * the 'email' property
+     *
+     * @param {Array} $contactIds
+     * @return {Array}
+     */
+    private static function _getContactsWithEmail($contactIds) {
+        $contacts = civicrm_api3('Contact', 'get', array(
+            'return' => 'display_name',
+            'id' => array('IN' => $contactIds),
+            'api.Email.getsingle' => array(
+                'contact_id' => "\$value.contact_id",
+                'is_primary' => true,
+                'return' => array('contact_id', 'email')
+            )
+        ))['values'];
+
+        foreach ($contacts as $id => $contact) {
+            $contacts[$id]['email'] = $contact['api.Email.getsingle']['email'];
+            unset($contacts[$id]['api.Email.getsingle']);
+        }
+
+        return $contacts;
+    }
+
+    /**
+     * Returns the links (in <a> tags), names and email of the contacts with given ids
+     * It also populates an array that maps emails to contact ids
+     *
+     * @param {Array} $contactIds
+     * @param {Array} $emailToContactId (by reference) Mapping between emails and contact ids
+     * @return {Array}
+     */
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
```diff
@@ -36,98 +36,129 @@ private static function _checkRelatedExtensions()
...
+    private static function _getContactsWithEmail($contactIds) {
+        $contacts = civicrm_api3('Contact', 'get', array(
+            'return' => 'display_name',
+            'id' => array('IN' => $contactIds),
+            'api.Email.getsingle' => array(
+                'contact_id' => "\$value.contact_id",
+                'is_primary' => true,
+                'return' => array('contact_id', 'email')
+            )
+        ))['values'];
```
nishant-bhorodia: There could be a php version compatibility issue here. Please can we change it like this:

```
$result = civicrm_api3('Contact', 'get', array(
    'return' => 'display_name',
    'id' => array('IN' => $contactIds),
    'api.Email.getsingle' => array(
    'contact_id' => "\$value.contact_id",
    'is_primary' => true,
    'return' => array('contact_id', 'email')
  )
));

$contacts = $result['values'];
```

AkA84: Do you have a specific version in mind where this syntax could be a problem?

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
```diff
@@ -36,98 +36,129 @@ private static function _checkRelatedExtensions()
...
     }
-    
+
+    /**
+     * Gets the contacts starting with the given ids
+     *
+     * It also chains a call to the 'Email' api and puts the value in
+     * the 'email' property
+     *
+     * @param {Array} $contactIds
```
nishant-bhorodia: I think the data type array is written as `array` rather than `{Array}` in the previous comments maybe ? Please have a look at `private static function _send`.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
nishant-bhorodia: I believe I've had this issue with PHP 5.3

nishant-bhorodia:
Thank you for the changes @AkA84 . Approved !

PR #42, opened 2016-04-08
PCHR-901: Task and Documents tab on Contact page are stuck in a loading loop

AkA84:
### Problem
The "Tasks" and "Documents" tabs in the Contact page weren't loading. The former appeared to be stuck in a loading loop, the latter just didn't show anything

before

### Solution
There were two issues that caused the bug:
1. We moved the t&a app from `ngRoute` to `ui.router` ([commit](https://github.com/compucorp/civihr-tasks-assignments/commit/770dea1112a177cf99599c21929af25a12fa17fd)) but forgot to change the `ng-view` directive into `ui-view` in the documents and tasks templates
2. The `textAngular` angular module was causing problems, which were fixed by using the version store in the `common/` dependencies and by [this PR](https://github.com/civicrm/civihr/pull/890)

after


kkalamarski:
Approved

PR #43, opened 2016-04-08
PCHR-637: Fix 'add to assignment' in documents modal.

kkalamarski:
Addressing the same issue as #35, but this time for 'add new document' modal.

## Description
Adding more informations (id, status, creation date, subject) to the Assignment lookup select dropdown, to make it work similar to the "File to Case" feature in default CiviCRM

### Before
drop_before

### After
dropdown_after


AkA84:
Approved

PR #41, opened 2016-04-08
PCHR-778: Add task design

kkalamarski:
## 1. Problem
Modal was too wide.
width_before

### Solution
Modal width was changed from `600px` to `500px` to meet the design.
width_after

## 2. Problem
Due date should be set to today's date by default (pre-filled).
date_before

### Solution
Due date is now set to today's date.
date_after

## 3. Problem
When a task is created, Target and Assignee fields are empty in the dashboard, unless you refresh the page.

**Reason:**
method `updateCache()` was altering cache with object what was lacking `display_name` field. Therefore template was trying to display field that no longer exist in model. 
missing_name_before

### Solution
`display_name` field was added to updateCache method. 
missing_name_after

PR #44, opened 2016-04-08
PCHR-779: Add New Document

kkalamarski:
# Problems addressed in this PR

## Problem
When expiry date is set to 19/01/2017 error is thrown: `custom_97 is not a valid date: 19/01/2017` after saving. 
Modal disappears but no new document is created.

**Reason of bug:** Problem existed when date was entered into date field using keyboard. Server assumes that date is sent in `mm/dd/yyyy` format, while it was sent in `dd/mm/yyyy` format. Because of that date `19/01/2017` wasn't read properly.

### Solution
Parse dates to `yyyy-mm-dd` format (the safest one, which can easily be read by server) before sending them to server.

## Problem
`expiry date` field is mandatory, but not all documents would have expiry date. 
Mandatory fields should be: `Target name`, `document type`, `due date` and `assignee`.
### Solution
Removed `ng-required` directive from `expiry date` field. Made sure that all mandatory fields have `ng-require`.

## Problem
Major design flaws.
- Date fields wasn't correctly displaying dates
- Files table was unreadable.
- Due Date label was split into 2 lines

**Before:**

table_before

**After:**

table_after

# Problems not addressed in this PR but pointed in the ticket.

## Problem
Add to assignment field does nothing.
### Solution
Fixed in #43.

##  Problem
Modal looks a lot wider than the mock.
### Solution
Fixed in #41.


uk.co.compucorp.civicrm.tasksassignments/scss/civitasks/partials/_forms.scss
```diff
@@ -1,10 +1,15 @@
 form {
 
+  /* helper classes */
   .no-gutter[class*='col-'] {
     padding-left: 0;
     padding-right: 0;
   }
 
+  .no-wrap {
```
AkA84: There is already a utility class in Bootstrap for this, it's `.text-nowrap`

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -209,6 +212,27 @@ define([
...
+             * @param {string|Date} date
+             * @returns {string|null}
+             */
+            $scope.parseDate = function (date) {
+
+                if (date instanceof Date) {
+                    date = date.getTime();
+                }
+
+                var formatted = moment(date, [
```
AkA84: Please leave a comment as to why you put this formats in the list

AkA84:
Couple of minor things to address, other than that it's approved

PR #45, opened 2016-04-11
PCHR-896: Point assignments tab to civicrm/case

AkA84:
## Problem
The Assignments route was set up inconsistently. On one hand it was assigned the `appraisals.html` template, which is just an iframe pointing to the `civicrm/case` page, whose url was available through the `ExternalPageCtrl` (the same ctrl used for the Reports route). On the other hand the controller assigned to the route wasn't `ExternalPageCtrl`, but `AssignmentsCtrl`, which is empty. End result was that the page didn't load anything

before

## Solution
Switching the route to use `ExternalPageCtrl` solves the issue

after


kkalamarski:
Approved

PR #46, opened 2016-04-11
PCHR-573: Assignment menu links need to be cleaned up

AkA84:
We need to disable the menu items of the Case extension (`org.civicrm.hrcase`) when Tasks&Assignments is enabled.
To be sure this change is retroactive, an upgrader has been written as well


coldrunKacper:
Approved.

PR #47, opened 2016-04-11
PCHR-908: Disable form submit when pressing enter on lookup dropdown

AkA84:
An old version of `ui-select` (0.11.2) was used, which caused an unwanted form submit to happen whenever the user pressed enter to select one of the lookup results in the dropdown

Using the newest version (0.14.8) available in the `common/` dependencies solves the problem


omarabuhussein:
approved

PR #48, opened 2016-04-18
PCHR-772: Remove local font-awesome partial + html fix

AkA84:
The tasks and assignments extensions was include in its own css the font-awesome stylesheet, but that's already provided by `the org.civicrm.bootstrapcivicrm extension`, so we can remove it safely and avoid specificity issues


kkalamarski:
Approved

PR #49, opened 2016-04-19
PCHR-772: Swap lookup directive for normal select dropdown

AkA84:
As a quickfix before the ui-select dropdown is properly customized, in case where the directive is used just as a standard select dropdown (that is, single value without the search/filter functionality enabled), then the directive is removed and a normal custom-style select element (`.crm_custom-select`) is used

Since the T&A extension is still not using the common style and since .crm_custom-select is defined in there, then as a quick fix the `.crm_custom-select` is redefined as a local partial in the extension (ugly fix, for we can't wait until we make the full swap to using the common style)

#### Before

before-task
before-assignment
before-document

#### After
after-task
after-assignment
after-document


danilo-compucorp:
Approved

PR #50, opened 2016-04-20
PCHR-983: Disable autocomplete caching for contact lookup in T&A

AkA84:
### Problem
There was a problem in the Document modal in "edit mode" which was the same as the problem addressed in #36 for the Task and Assignment modals: not all the data of an already existing document was displayed correctly

before_document_selected

### Solution
Copying it straight from the aforementioned PR:
> 1. The ng-show logic wasn't implemented correctly, that is the fields weren't shown if the data was already there
2. For the lookup lists loaded via ajax (assignments, assignees, target), the list given to the lookup directive must come pre-loaded with the item already selected for the task

after_document_selected

---

#### Problem
The lookup directives in all 3 modals (task, assignment, document) were sharing the same result collection , which resulted in the users getting a list of results before he even typed something in the search field

before_document_cache
before_assignment
before_task

#### Solution
Splitting the `$scope.contacts` collection into separate subsets (one for each directive) make sure that the list of results of one directive istance does not leak into another

after_document_cache
after_assignment
after_task


danilo-compucorp:
Approved

PR #52, opened 2016-04-22
PCHR-994: Show Target/Assignee name after edit-in-place change

AkA84:
### Problem
When assigning, via edit-in-place, a target or an assignee to a task which hasn't been already cached (that is, wasn't already assigned to some other task in the list), the name of the contact wouldn't appear after the change had been submitted

before

There were several problems here:
1) There was a `$scope.cacheContact` method that was called on the `on-select` listener of the `editable-ui-select` directive, but that was never called. I'm not sure yet if that's a problem with the `ui-select` directive itself or with its xeditable version. Even if it worked it still wouldn't have been right, as the selected, not the submitted, contact would've been cached, while we want to cache only the submitted contact

2) The xeditable version of `ui-select` doesn't work with the `repeat="contact in contacts"` syntax, as it works only when you specify a property of the tracked object (`repeat="contact.id as contact in contacts"`). This means that we can only work with the id of the contact if we want to pass it to `cacheContact`, and not the whole object

### Solution
I've remove the `on-select` listener since it wasn't working and added the call to `cacheContact` on the `onbeforesave` listener, acting as a guard before the task is updated

after

Also i've amended the implementation of `cacheContact` itself, so that now it accepts a contact id (the only data we can get from the `ui-select` directive) which it uses to find the about-to-be-submitted contact in the list of cached contacts (the one that the `ui-select` directive uses to generate the dropdown list)


danilo-compucorp:
Looks ok. PR approved.

PR #54, opened 2016-05-07
PCHR-478: Fix assignees' bulk assignment

AkA84:
### Problem
After the assignees have been bulked-assigned the first time to the whole list of tasks/documents, attempting to do it again failed for those tasks/documents that in the meantime got a diffent assignee manually assigned to them

before

This was due to a couple of reasons:
1. The ui-select ng-model value can switch to a new value only if it is among the full list of items that have been fed to the directive
2. Both the task and document items shared two common list of contacts (`$scope.contacts.task` and `$scope.contacts.document`) for their ui-selects. This meant that searching a new assignee for task no.2 meant that the lookup contact list of task no.5 changed as well

As a result once one of the tasks/documents had been manually assigned to an assignee (which meant fetching a new contacts list), if the common assignee wasn't among the new contacts list nothing would get displayed

### Solution
The `task` and `document` lists are now split into sub-lists, one for each item, so that the various ui-select cannot affect one another. Also this means that when the bulk assignment is done, the contacts sub-list of the first item is simply copied to the other items, and the common assignee is correctly displayed

after


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
```diff
@@ -79,15 +82,32 @@ define([
                 $scope.dpOpened[key] = true;
             };
 
-            $scope.refreshContacts = function (input, type) {
+            /**
+             * Fetches and stores a list of contacts based on the given
+             * query string
+             *
+             * @param {string} input - The search query the user typed
```
danilo-compucorp: To be more consistent, I'd put the description in a new line, just like the following param descriptions.

danilo-compucorp:
Looks ok. PR approved.

PR #57, opened 2016-05-11
PCHR-1036: Hide "Follow Link" for the initial launch of T&A

danilo-compucorp:
This PR removes the "Follow Link" button from Tasks and Assignments (both target and assignee).
The old behavior was:
pchr-1036-before
 
And the new behavior is:
pchr-1036-after


brunoqueiros:
Approved!

PR #64, opened 2016-06-27
PCHR-1075: Fix permissions

AkA84:
#### Problem 
The "Tasks and Assignments" main menu item and the "Tasks" and "Assignments" tabs in the contact page were visible and accessible even if the user didn't have the "access Tasks and Assignments" credentials assigned to it

Also, the `access_arguments` values weren't specific enough for some of the pages defined in `tasksassignments.xml`

### Solution
In the `tasksassignments_civicrm_tabs` hook, the tabs are added only if the user has the credential
```php
function tasksassignments_civicrm_tabs(&$tabs) {
  if (CRM_Core_Permission::check('access Tasks and Assignments')) {
    $tabs[] = Array(/* ... */);
  }
```
The upgrader function that adds the T&A main menu item (`upgrade_0002`) has been amended so that it sets the permission correctly on it, but to be sure that the change is applied retroactively, a new upgrader has also been added which enforces the permission

The `taskassignments.xml` has been refactored so that the pages are better grouped, and some of the pages have now more specific restrictions

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -1,7 +1,7 @@
 <?php
 
 class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
-{    
+{
     public function install() {
 
       // $this->executeCustomDataFile('xml/customdata.xml');
@@ -14,13 +14,13 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
             $this->{$methodName}();
           }
       }
-      
+
     $this->setComponentStatuses(array(
       'CiviTask' => true,
       'CiviDocument' => true,
     ));
     }
-    
+
   public function enable() {
     $this->setComponentStatuses(array(
       'CiviTask' => true,
@@ -28,14 +28,14 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
     ));
     return TRUE;
   }
-  
+
   public function disable() {
     $this->setComponentStatuses(array(
       'CiviTask' => false,
       'CiviDocument' => false,
     ));
   }
-    
+
   /**
    * Set components as enabled or disabled. Leave any other
    * components unmodified.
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -468,21 +468,48 @@ public function upgrade_1017()
...
+    $taNavigation = new CRM_Core_BAO_Navigation();
+    $taNavigation->name = 'tasksassignments';
+    $taNavigation->find(true);
+
+    if($taNavigation->id && !$taNavigation->permission) {
+      $navigation = new CRM_Core_BAO_Navigation();
+      $foo = array(
+        'id' => $taNavigation->id,
+        'permission' => 'access Tasks and Assignments',
+        'separator'  => 1,
```
omarabuhussein: I think there is no need for **separator** and **is_active** to be updated

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -468,21 +468,48 @@ public function upgrade_1017()
...
     return false;
   }
-    
+
+  /**
+   * Make sure that the Tasks and Assignments main menu item
+   * (and subsequently the Dashboard menu item) are restricted to only the users
+   * with 'access Tasks and Assignments' permission
+   */
+  public function upgrade_1018() {
```
omarabuhussein: we could update the current civihr setups manually ... so I think there is no need for this new upgrader .. what do you think ?

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -173,25 +173,30 @@ function tasksassignments_civicrm_pageRun($page) {
...
-    if ($documentsTab['values']['documents_tab']['value']) {
-        $tabs[] = Array(
-            'id'        => 'cividocuments',
-            'url'       => CRM_Utils_System::url('civicrm/contact/view/documents'),
-            'title'     => ts('Documents'),
-            'weight'    => 50
-        );
-    }
 
+    if (
```
omarabuhussein: a single check for permission is enough ... so you could change the code to something like :

if (CRM_Core_Permission::check('access Tasks and Assignments')) {
      $tabs[] = Array(
        'id'        => 'civitasks',
        'url'       => CRM_Utils_System::url('civicrm/contact/view/tasks'),
        'title'     => ts('Tasks'),
        'weight'    => 40
      );
    $documentsTab = civicrm_api3('TASettings', 'get', array(
      'fields' => 'documents_tab',
    ));

if ( $documentsTab['values']['documents_tab']['value']) {
      $tabs[] = Array(
        'id'        => 'cividocuments',
        'url'       => CRM_Utils_System::url('civicrm/contact/view/documents'),
        'title'     => ts('Documents'),
        'weight'    => 50
      );
    }
    }

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
AkA84: Unfortunately there is, if you don't define them (in 4.5) it defaults them to `false`, see [here](https://github.com/civicrm/civicrm-core/blob/4.5/CRM/Core/BAO/Navigation.php#L94-L95)
```php
$params['is_active'] = CRM_Utils_Array::value('is_active', $params, FALSE);
$params['has_separator'] = CRM_Utils_Array::value('has_separator', $params, FALSE);
```

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
AkA84: I would err on the side of caution, tbh. You never know.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
AkA84: You're right, sloppy logic on my part, i'll amend

omarabuhussein:
looks good .. approved :+1:

PR #65, opened 2016-06-28
PCHR-1160 - Making 'source_contact_id', 'target_contact_id' and 'assignee_contact_id' required fields when creating Task / Document on BAO level.

coldrunKacper:
Task / Document entities were handling creation logic the same way as original Activity entity so these three contact fields were not required on API level when creating an entity instance. I've modified both BAOs to require these contacts IDs when creating a new entity instance.

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
@@ -12,6 +12,14 @@ class CRM_Tasksassignments_BAO_Document extends CRM_Tasksassignments_DAO_Documen
     {
         $entityName = 'Document';
         $hook = empty($params['id']) ? 'create' : 'edit';
+
+        if ($hook === 'create') {
+            // If creating a new Document, we require all three contacts to be defined.
+            if (empty($params['source_contact_id'] || empty($params['target_contact_id']) || empty($params['assignee_contact_id']))) {
+                throw new CiviCRM_API3_Exception("Please specify 'source_contact_id', 'target_contact_id' and 'assignee_contact_id'.");
+            }
+        }
+
         CRM_Utils_Hook::pre($hook, $entityName, CRM_Utils_Array::value('id', $params), $params);
         
         if (empty($params['status_id']) && $hook === 'create')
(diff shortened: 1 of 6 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -12,6 +12,14 @@ public static function create(&$params)
     {
         $entityName = 'Document';
         $hook = empty($params['id']) ? 'create' : 'edit';
+
+        if ($hook === 'create') {
+            // If creating a new Document, we require all three contacts to be defined.
+            if (empty($params['source_contact_id'] || empty($params['target_contact_id']) || empty($params['assignee_contact_id']))) {
+                throw new CiviCRM_API3_Exception("Please specify 'source_contact_id', 'target_contact_id' and 'assignee_contact_id'.");
```
omarabuhussein: change :
> Please specify 'source_contact_id', 'target_contact_id' and 'assignee_contact_id'.

to 

Please specify 'source_contact_id', 'target_contact_id' **or** 'assignee_contact_id'.

and do the same in the other BAO file. But I think that the message should be more specific to what is really missing .

omarabuhussein: Why throughing  CiviCRM_API3_Exception ... The API is not the only place that call this method ... as you know you can write :
`CRM_Tasksassignments_BAO_Document::create($params)` in any place in the so throwing API exception doesn't make much sense .
I think there is CiviCRM_Error_Exception or something like that could be used here.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -12,6 +12,14 @@ public static function create(&$params)
...
+            }
```
omarabuhussein: If we forget what I said above for a while ... you can refactor the code to this :

```php
if ($hook === 'create') {
  // If creating a new Document, we require all three contacts to be defined.
  if (empty($params['source_contact_id'] || empty($params['target_contact_id']) || empty($params['assignee_contact_id']))) {
    throw new CiviCRM_API3_Exception("Please specify 'source_contact_id', 'target_contact_id' and 'assignee_contact_id'.");
  }
  if (empty($params['status_id']))
  {
    $params['status_id'] = 1;
  }
}

 CRM_Utils_Hook::pre($hook, $entityName, CRM_Utils_Array::value('id', $params), $params);
```

same apply for /Tasksassignments/BAO/Task.php

omarabuhussein:
Looks good .. approved

PR #66, opened 2016-06-29
PCHR-1140: Fix edit-in-place

AkA84:
#### Problem
When trying to use the edit-in-place to change the details of a given task, an error would appear

before

This was due to the fact that, when updating, only the id of the task and the data that was changed by the user was passed to the service responsible for updating the record. Given that important data related to the task (like `case_id`, or `activity_type_id`) was missing, the backend would throw an error

#### Solution
Passing the whole task (not by reference) to the service solved the problem

after


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/task-list.js
```diff
@@ -210,16 +210,11 @@ define([
...
-                var updatePromise = TaskService.save(angular.extend({
-                    id: task.id
-                },updateObj));
-
-                updatePromise.then(null,function(reason){
-                    CRM.alert(reason, 'Error', 'error');
+              return TaskService
+                .save(angular.extend({}, task, updateObj))
+                .catch(function (reason) {
+                  CRM.alert(reason, 'Error', 'error');
```
deb1990: Instead the having same "Error" twice, can we mention method name in the 2nd/last parameter?

AkA84: Can you give me an example of what you think it should look like?

deb1990: Something like, CRM.alert(reason, '<MethodName>', 'Class/FileName');

deb1990: Nevermind, as this an error visible to user, its better not to show method name

deb1990:
Approved

PR #68, opened 2016-07-01
PCHR-1174: Document - Clicking on Contact name takes you to different…

danilo-compucorp:
Issue:
In Tasks and Assignments / Documents page, clicking on a Contact name was taking users to incorrect contact's page.

Example: 
Clicking on "krish, madu" should go to CID=220, but instead it was pointing to CID=67

screen shot 2016-06-09 at 14 05 39

The issue was located in the template file (documents.html), where the link's url was incorrectly being created using the "id" property, instead of the expected "contact_id".

To better understand it, below is an example of "cache.contact.obj" state, where the issue can be seeing:
cache


AkA84:
Missing explanation of cause and fix. After that is added the PR is approved

PR #69, opened 2016-07-04
PCHR-1191: Remove joining & exiting case types from T&A extension

omarabuhussein:
## Problem
Both **hrcase** and **Tasks & Assignments** Extensions try to create the following case types ( Joining , Exiting ) , but since there is a unique constraint on case type name field in civicrm_case_type table this is not possible and many errors arisen which is reported here  PCHR-1151 and later on here  PCHR-1191 .

## Solution

 After starting a discussion on the later one we decided to remove these activity types from **Tasks & Assignments** extension since these case types are related to HR , and in case **Tasks & Assignments** is used with civihr then hrcase have to be enabled too.


coldrunKacper:
Approved.

omarabuhussein:
replaced by :  https://github.com/compucorp/civihr-tasks-assignments/pull/70

PR #73, opened 2016-07-07
PCHR-1277: Fix AngularJS tests in CiviCRM 4.7

AkA84:
See https://github.com/civicrm/civihr/pull/1004


brunoqueiros:
Looks good, approved!

PR #75, opened 2016-07-12
PCHR-1180: Filtering documents by the due date only

danilo-compucorp:
The issue was that the same document was listed in both _Overdue_ and _Due this Week_ filters. The _Due date_ was falling under the _Overdue_ category and _Expiry Date_ was falling under _Due this week_ category. This should not happen, since the numbers highlighted in filters then don’t match the actual number of documents listed.
This PR makes the document filtering only consider the _due date_ field.


AkA84:
Approved

PR #76, opened 2016-07-14
PCHR-1183: Fixing the status for the document emails

danilo-compucorp:
The status present in the emails were incorrect for the document notifications.
This PR fixes those status. 

Before: 
before

After:

Documents:
documents

Tasks:
tasks


coldrunKacper:
Approved.

PR #77, opened 2016-07-15
PCHR-1226: Showing the "status" field only after "set status" is …

danilo-compucorp:
This PR makes the _status_ field only appear when the _set status_ is clicked.

Before:
before

After:
after


AkA84:
Approved

PR #79, opened 2016-07-20
PCHR-1263: ensure hrcase dependency on T&A

omarabuhussein:
Linked PR : https://github.com/civicrm/civihr/pull/1020

## Problem
The user should not be able to uninstall/disable T&A in case hrcase extenstion is enabled.

## Solution

I've checked if hrcase is installed and enabled inside both  **tasksassignments_civicrm_uninstall** and **tasksassignments_civicrm_disable** hooks so if it is  installed and enabled the user will be redirected to extensions manager page and an error notification will appear informing the user to uninstall/disable hrcase extension first .

anim


coldrunKacper:
Approved.

PR #82, opened 2016-07-28
PCHR-1182: Changing the upload permission

danilo-compucorp:
This PR fixes the permission required to upload files from the document block on the SSP.

**Issue**: 
As civihr_staff  
From Document block > Upload button, you can select file, but Save button does nothing and unable to upload document

**Solution**:
Currently, to upload files, the "access Tasks and Assignments" permission is required (which the civihr_staff doesn't have). 
This PR changes it to "access uploaded files" permission (which the civihr_staff has).


coldrunKacper:
Approved.

PR #85, opened 2016-08-09
PCHR-1410: Update info.xml file

AkA84:
Updating the info.xml file with the lastest value for the `<releaseDate>`,  `<version>`, and `<ver>` fields


omarabuhussein:
if you are sure this should be marked as  "beta" then approved.

PR #88, opened 2016-08-16
PCHR-1361: Replace assignee with target in documents list

AkA84:
Instead of showing the assignee contact on the document list, we now show the target contact
_(Also: the minified js file apparently needed to be updated)_

**Before**
before

**After**
after


kkalamarski:
Approved

PR #89, opened 2016-08-17
PCHR-1364: Delete dummy document types and add real ones

AkA84:
We were using dummy data for the document types values:
* Joining Document 1
* Exiting Document 1

They are not being replaced by real, default values:
* VISA
* Passport
* Government Photo ID
* Driving licence
* Identity card
* Certificate of sponsorship (COS)


nishant-bhorodia:
Approved!

omarabuhussein:
In general I don't agree in the way upgrader is used , the upgrader should be only used for old installation , for new installations stuff should happen by updating *_install.sql file . but since how it's done currently I will pass it but it need to be refactored later. ( for more info please read **"Execute database upgrades during extension installation**"  section at this page https://compucorp.atlassian.net/wiki/pages/viewpage.action?pageId=80478227

PR #90, opened 2016-08-17
PCHR-1477: Remove the "Assignments" tab and the "Cases" top nav item

AkA84:
**Before**
before

**After**
after

### Update
Also now the T&A extension, when enabled, will *always* remove the "Cases" main menu item, while before it was done only when `hrcase` was already enabled ("Cases" does not really belong to `hrcase`, but rather to CiviCRM in general, so there's no use in making that distinction based on `hrcase`).

The original upgrader was renamed as to be the newest  (i tested it and it doesn't seem to create any problem the fact that there is a "hole" between upgrades)

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -438,26 +438,6 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
   }
 
   /**
-   * Disables the Case menu items if Tasks&Assignments is enabled
-   *
-   * @return {boolean}
-   */
-  public function upgrade_1017()
-  {
-    $isEnabled = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_Extension', 'uk.co.compucorp.civicrm.tasksassignments', 'is_active', 'full_name');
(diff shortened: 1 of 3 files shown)


omarabuhussein:
approved

omarabuhussein:
can you rename _toggleCaseMenuItems to _ta_toggleCaseMenuItems

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -484,6 +464,22 @@ public function upgrade_1018() {
     return false;
   }
 
+  /**
+   * Disables the Case menu items if Tasks&Assignments is enabled
+   *
+   * @return {boolean}
+   */
+  public function upgrade_1019() {
```
omarabuhussein: why did you moved the logic to 1019 , isn't this 1017 already executed and I see no difference between 1017 and 1019

AkA84: the difference is that we disabled "Cases" no matter if {{hrcase}} is enabled. I moved it to 1019 so that we didn't have basically two duplicate upgraders, and still the upgrade would run on both new and already-installed extensions

omarabuhussein: oh got it ...thanks

omarabuhussein:
approved

PR #91, opened 2016-08-22
PCHR-821: Fix filters in tasks dashboard

AkA84:
## Problem
In the Tasks dashboard, the "All" top filter wasn't working correctly: instead of showing literally all the unresolved tasks currently existing in the system, it was showing only the sum of the "My" and "Delegated" lists

before

## Solution
The page now loads the all the unresolved tasks so that "All" can display the correct list

after

#### Optimization
Loading all the tasks it's not obviously optimized for systems that in the long run will have many tasks, although this loads only the _unresolved_ ones, so their number in theory shouldn't be very high. Nevertheless this whole initial solution could benefit from a whole different approach, but given time constraints I decided to limit the scope of this PR

#### `filterByOwnership` filter
The `filterByUserRole` has now being replaced by the new `filterByOwnership` filter. The former had a signature that I didn't like much, given that you had to specify the name of a field plus a boolean flag. The new one instead encapsulates the implementation details within it, only requiring a string parameter to function

#### `sidebar-filters` directive
The sidebar filters in the task and in the documents dashboard were basically the same, except for an additional filter in the dashboards one:

sidebar-tasks sidebar-documents

So they were unified in a single directive: `sidebar-filter`. The directive is kept simple on purpose (to avoid the rabbit-hole effect), having a shared scope and transclude tag that allows the documents dashboard to insert the additional field. 

Additionally, since the scope is shared in the directive, the two controllers of the two dashboards had to use the same variable names for their list of task/documents. So `taskList` and `documentList` became both `list`, `taskListOngoing` and `documentListOngoing` became both `listOngoing`, etc


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/document-list.js
```diff
@@ -113,18 +113,15 @@ define([
...
-                    field: null,
-                    isEqual: null
-                },
-                dateRange: {
-                    from: null,
-                    until: null
-                },
-                due: null,
-                assignmentType: []
+              contactId: null,
```
deb1990: Since document-list.js and task-list.js using same $scope.filterParams. We should fetch it from a function to reduce redundancy. It can be from a service/controller function/constant.

AkA84: Isn't a service an overkill for a small json object? Not to mention that there is at least one difference between one `filterParams` and the other, which in the future could become two or more, undermining the reason for the service to exist in the first place. Could you provide an example of what you have in mind?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/modules/app-dashboard.js
```diff
@@ -202,37 +161,24 @@ define([
...
+                                  'assignee_contact_id': config.LOGGED_IN_CONTACT_ID,
+                                  'status_id': { 'NOT IN': config.status.resolve.TASK }
+                                }),
+                                TaskService.get({
+                                  'sequential': 0,
+                                  'source_contact_id': config.LOGGED_IN_CONTACT_ID,
+                                  'status_id': { 'NOT IN': config.status.resolve.TASK }
+                                })
+                              ]).then(function (results) {
+                                deferred.resolve(_(results[0]).assign(results[1]).values().value());
```
deb1990: Can we simplify this part?
`_(results[0]).assign(results[1]).values().value()` 
Something like;
`var value = results[1]).values().value();
_(results[0]).assign(value)`

Just to make things more readable.

uk.co.compucorp.civicrm.tasksassignments/views/dashboard/documents.html
```diff
@@ -6,9 +6,15 @@ <h3 class="panel-title">View Documents By:</h3>
...
-                    <li role="presentation"><a href ng-model="filterParams.due"  uib-btn-radio="'dueThisWeek'" uncheckable>{{label.dueThisWeek}} <i class="fa fa-calendar" ng-click="viewInCalendar('week');"></i> <span class="badge badge-primary" ng-bind="(documentListOngoing | filterByDue:'dueThisWeek' | filterByUserRole:filterParams.userRole.field:filterParams.userRole.isEqual).length"></span></a></li>
+                  <li role="presentation">
+                    <a href ng-model="filterParams.due" uib-btn-radio="'overdue'" uncheckable>{{label.overdue}} <i class="fa fa-calendar" ng-click="viewInCalendar('month');"></i> <span class="badge badge-danger" ng-bind="(documentListOngoing | filterByDue:'overdue' | filterByOwnership:filterParams.ownership).length"></span></a>
+                  </li>
+                  <li role="presentation">
+                    <a href ng-model="filterParams.due" uib-btn-radio="'dueToday'" uncheckable>{{label.dueToday}} <i class="fa fa-calendar" ng-click="viewInCalendar('day');"></i> <span class="badge badge-primary" ng-bind="(documentListOngoing | filterByDue:'dueToday' | filterByOwnership:filterParams.ownership).length"></span></a>
```
deb1990: Please change the indentation just like the below to make it more readable,
<a 
href ng-model="filterParams.due" 
uib-btn-radio="'dueToday'"
.....

deb1990: Also the span should be in a new line

uk.co.compucorp.civicrm.tasksassignments/views/dashboard/documents.html
```diff
@@ -6,9 +6,15 @@ <h3 class="panel-title">View Documents By:</h3>
...
-                    <li role="presentation"><a href ng-model="filterParams.due" uib-btn-radio="'dueToday'" uncheckable>{{label.dueToday}} <i class="fa fa-calendar" ng-click="viewInCalendar('day');"></i> <span class="badge badge-primary" ng-bind="(documentListOngoing | filterByDue:'dueToday' | filterByUserRole:filterParams.userRole.field:filterParams.userRole.isEqual).length"></span></a></li>
-                    <li role="presentation"><a href ng-model="filterParams.due"  uib-btn-radio="'dueThisWeek'" uncheckable>{{label.dueThisWeek}} <i class="fa fa-calendar" ng-click="viewInCalendar('week');"></i> <span class="badge badge-primary" ng-bind="(documentListOngoing | filterByDue:'dueThisWeek' | filterByUserRole:filterParams.userRole.field:filterParams.userRole.isEqual).length"></span></a></li>
+                  <li role="presentation">
+                    <a href ng-model="filterParams.due" uib-btn-radio="'overdue'" uncheckable>{{label.overdue}} <i class="fa fa-calendar" ng-click="viewInCalendar('month');"></i> <span class="badge badge-danger" ng-bind="(documentListOngoing | filterByDue:'overdue' | filterByOwnership:filterParams.ownership).length"></span></a>
```
deb1990: Same as below

deb1990: We can create a array in controller like
`$scope.filterConfig = [
  {
    "radio": "overdue",
    "icon": "calender",
    "calenderType": "month",
    "badge" : "badge badge-danger",
  },...
]`

and then use ng-repeat to reduce the code.

uk.co.compucorp.civicrm.tasksassignments/views/dashboard/documents.html
```diff
@@ -78,27 +84,30 @@ <h3 class="panel-title">{{label[filterParams.due]}}</h3>
...
-                                   isEqual: false
-                               }">Delegated Docs ({{(documentListOngoing | filterByUserRole:'assignee_contact_id':false).length}})</a>
-                            <a ui-sref="documents.all"
-                               class="btn btn-gray panel-filter__filter"
-                               ng-model="filterParams.userRole"
-                               uib-btn-radio="{
-                                   field: null,
-                                   isEqual: null
-                               }">All</a>
+                          <a
```
deb1990: Similarly can we create a array and use ng-repeat here too.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/document-list.js
deb1990: Sorry did not notice the difference in "documentStatus". We can keep it.

uk.co.compucorp.civicrm.tasksassignments/views/dashboard/documents.html
```diff
@@ -6,9 +6,27 @@ <h3 class="panel-title">View Documents By:</h3>
...
-                    <li role="presentation"><a href ng-model="filterParams.due" uib-btn-radio="'dueToday'" uncheckable>{{label.dueToday}} <i class="fa fa-calendar" ng-click="viewInCalendar('day');"></i> <span class="badge badge-primary" ng-bind="(documentListOngoing | filterByDue:'dueToday' | filterByUserRole:filterParams.userRole.field:filterParams.userRole.isEqual).length"></span></a></li>
-                    <li role="presentation"><a href ng-model="filterParams.due"  uib-btn-radio="'dueThisWeek'" uncheckable>{{label.dueThisWeek}} <i class="fa fa-calendar" ng-click="viewInCalendar('week');"></i> <span class="badge badge-primary" ng-bind="(documentListOngoing | filterByDue:'dueThisWeek' | filterByUserRole:filterParams.userRole.field:filterParams.userRole.isEqual).length"></span></a></li>
+                  <li role="presentation">
```
deb1990: There is lot of common markup between documents.html and tasks.html which can be reduced using ng-include.

uk.co.compucorp.civicrm.tasksassignments/views/dashboard/documents.html
```diff
@@ -6,9 +6,27 @@ <h3 class="panel-title">View Documents By:</h3>
...
+                  </li>
+                  <li role="presentation">
+                    <a href ng-model="filterParams.due" uib-btn-radio="'dueToday'" uncheckable>
+                      {{label.dueToday}}
+                      <i class="fa fa-calendar" ng-click="viewInCalendar('day');"></i>
+                      <span class="badge badge-primary" ng-bind="(documentListOngoing | filterByDue:'dueToday' | filterByOwnership:filterParams.ownership).length"></span>
+                    </a>
+                  </li>
+                  <li role="presentation">
+                    <a href ng-model="filterParams.due"  uib-btn-radio="'dueThisWeek'" uncheckable>
```
deb1990: Can we place every attribute for the spans in a new line just like it is done for the below "a" tags which makes it more readable.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/task-list.js
```diff
@@ -81,12 +81,12 @@ define([
             $scope.dueToday = 0;
             $scope.dueThisWeek = 0;
             $scope.overdue = 0;
-            $scope.taskList = taskList;
-            $scope.taskListFiltered = [];
-            $scope.taskListLimit = 5;
-            $scope.taskListOngoing = [];
-            $scope.taskListResolved = [];
-            $scope.taskListResolvedLoaded = false;
+            $scope.list = taskList;
```
deb1990: There is lot of common code between task-list.js and document-list.js. 
https://www.diffchecker.com/ry85ypxg
We can create a parent controller to reduce redundancy.

uk.co.compucorp.civicrm.tasksassignments/views/directives/sidebar-filters.html
```diff
@@ -0,0 +1,73 @@
+<div class="{{prefix}}filter-date">
+  <ul class="nav nav-pills nav-stacked">
+    <li role="presentation">
+      <a href ng-model="filterParams.due" uib-btn-radio="'overdue'" uncheckable>
```
deb1990: Can we do a ng-repeat here as I mentioned in my previous comment?

uk.co.compucorp.civicrm.tasksassignments/views/dashboard/tasks.html
```diff
@@ -91,34 +19,38 @@ <h3 class="panel-title">View Tasks By:</h3>
...
-                                    <a ui-sref="tasks.all"
-                                       class="btn btn-gray panel-filter__filter"
-                                       ng-model="filterParams.userRole"
-                                       uib-btn-radio="{
-                                            field: null,
-                                            isEqual: null
-                                        }">All
-                                    </a>
-                                </div>
+                              <div class="btn-group btn-group-sm">
```
deb1990: Can we do ng-repeat here?

deb1990:
Approved

PR #92, opened 2016-08-23
PCHR-1476: Remove top padding

brunoqueiros:
This PR remove the top padding from `#ct-dashboard` selector, and it was reverted `_config.scss` file to be able to generate CSS.

***Before***
image

***After***
image


AkA84:
Approved

PR #93, opened 2016-08-24
PCHR-1365: Create new document before original expiry date

coldrunKacper:
- Added a new 'days_to_create_a_document_clone' T&A setting field which keeps a number of days to create a document clone before its expire_date. If the value is '0' then no action is taken.
ta_clone_document_settings
- Added a scheduled job (daily) which checks for documents meeting the assumptions (approved and having 'expire_date' in X days) and then creates a clone of them.
ta_clone_documents_scheduled_job
- Added BAO methods to handle documents cloning.
- Added API entry point ('clonedocuments' action on Document entity).
- Added php unit tests for cloning documents flow.


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -1,35 +1,136 @@
...
+      'fields' => "days_to_create_a_document_clone",
+    ));
+    return !empty($setting['value']) ? $setting['value'] : 0;
+  }
+
+  /**
+   * Search for approved Documents with given expire date and create their clone.
+   * Return a count of cloned documents.
+   * 
+   * @param string $expireDate (date in Y-m-d format)
```
davialexandre: Recently, I've been thinking that, in cases like this, it's better if we enforce the use of DateTime objects instead of expecting a string with a date in the right format. Just like I did here: https://github.com/civicrm/civihr/pull/1104/commits/a0dd0dfc716d50a7f768c39d03f48475310c321c#diff-e17c168c0f4c4c52b05295b72cca13f6R281

These are the advantages I see on this approach:
1. The format gets irrelevant, which is good! Imagine that, for some reason, you need to change the format to Ymd instead of Y-m-d. Without a DateTime, you would need to change the dates on every call to this method
2. Implicit validation. You don't need to do date validation on your code to check if you received a valid date (Speaking of that, you didn't do it for this method)

Also, on cloneDocuments() (the method that calls cloneApprovedDocumentsWithExpireDate()) you already have a DateTime object

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -1,35 +1,136 @@
...
+      . 'WHERE DATE(ac.expire_date) = %1 AND a.status_id = 3';
+    $params = array(
+      1 => array($expireDate, 'String'),
+    );
+    $result = CRM_Core_DAO::executeQuery($query, $params);
+    while ($result->fetch()) {
+      $document = civicrm_api3('Document', 'getsingle', array(
+        'id' => $result->entity_id,
+        'return' => 'activity_type_id, activity_date_time, details, status_id, is_deleted, expire_date, case_id, source_contact_id, target_contact_id, assignee_contact_id',
+      ));
```
davialexandre: Why not joining this with $query and get everything with a single SQL query?

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/Api/DocumentTest.php
```diff
@@ -66,4 +71,80 @@ function testCreateDocumentWithNoTarget() {
...
+      'details' => $details2,
+      'activity_date_time' => $dueDate,
+      'expire_date' => $expireDate2,
+      'status_id' => 3,
+    ));
+
+    // Calling cloneDocuments() function (simulate scheduled job which runs daily).
+    $clonedCount = civicrm_api3('Document', 'clonedocuments');
+    // Now we should have one Document ('Some test Document 1') cloned (as
+    // its expire_date equals current date + 'days_to_create_a_document_clone'
```
davialexandre: Shouldn't this be "current date - 'days_to_create_a_document_clone'"?

coldrunKacper: No, it's fine: 
2016-08-24 (current date) + 5 days = 2016-08-29 - which is the 'expire_date' that we search for.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -1,35 +1,139 @@
...
+   * @return int
+   */
+  protected static function cloneApprovedDocumentsWithExpireDate($expireDate) {
+    $count = 0;
+    $query = "SELECT ac.id, ac.entity_id FROM civicrm_value_activity_custom_fields_11 ac "
+      . "LEFT JOIN civicrm_component c ON c.name = 'CiviDocument' "
+      . "LEFT JOIN civicrm_option_group og ON og.name = 'activity_type' "
+      . "LEFT JOIN civicrm_option_value ov ON ov.option_group_id = og.id AND ov.component_id = c.id "
+      . "LEFT JOIN civicrm_activity a ON a.id = ac.entity_id "
+      . "WHERE DATE(ac.expire_date) = %1 AND a.is_deleted = 0 AND a.status_id = 3 AND ov.component_id = c.id AND a.activity_type_id = ov.value";
```
davialexandre: What happens if, for some unknown/unpredicted reason the scheduled job doesn't get executed for one (or more days)? It seems the documents will never be cloned, right?

Example: I have a document where the expire date is 5 days ago and days_to_create_a_document_clone is 5. Then, for some reason, the scheduled job doesn't run today. When it runs tomorrow (or any other day after today), the query won't match that document and it will never be cloned. Is this a problem or is something expected?

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/Api/DocumentTest.php
```diff
@@ -9,6 +9,11 @@
 class Api_DocumentTest extends CiviUnitTestCase {
```
davialexandre: I know this not part of this ticket/PR, but the correct location of this file should be tests/phpunit/api/v3/DocumentTest.php and the class name should be api_v3_DocumentTest. 

If you could fix that it would be great!

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/Api/DocumentTest.php
davialexandre: Oh, right! You're adding days to get the expireDate:
`(new DateTime())->add(new DateInterval('P5D'))->format('Y-m-d');`

For some reason, I thought you were subtracting. Sorry!

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
coldrunKacper: It's more clean and handy (I guess) as we use the $document result for 'create' API call in the very next step (with minor changes to some properties). I guess getting the Document directly from SQL query (with another JOINs) would require additional formatting its values as Document 'get' call isn't just pure entry point for BAO get method. It's made basing on Activity entity and has some crazy logic inside.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
coldrunKacper: Yes, right. I'll ask Mike about this case (do we need to clone documents having expire date in X days or less).

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -78,7 +79,7 @@ protected static function cloneApprovedDocumentsWithExpireDate($expireDate) {
       . "LEFT JOIN civicrm_option_group og ON og.name = 'activity_type' "
       . "LEFT JOIN civicrm_option_value ov ON ov.option_group_id = og.id AND ov.component_id = c.id "
       . "LEFT JOIN civicrm_activity a ON a.id = ac.entity_id "
-      . "WHERE DATE(ac.expire_date) = %1 AND a.is_deleted = 0 AND a.status_id = 3 AND ov.component_id = c.id AND a.activity_type_id = ov.value";
+      . "WHERE DATE(ac.expire_date) >= CURDATE() AND DATE(ac.expire_date) <= %1 AND ac.clone_date IS NULL AND a.is_deleted = 0 AND a.status_id = 3 AND ov.component_id = c.id AND a.activity_type_id = ov.value";
```
davialexandre: Can you please add some tests to add coverage to this change? I can think of a few scenarios that should be tested:
- Test that a document isn't cloned more than once after consecutive calls to the scheduled job
- Test that the clone_date is set properly (You don't really need this if the previous test is working, but it's ok to add it if you feel like doing it)
- Test that the scheduled job will clone anything <= than the current date + 'days_to_create_a_document'

davialexandre:
Everything looks good. Just need one minor fix to the testCreateDocumentClone() test. It needs to be updated to reflect the latest changes to the test, saying it creates 3 documents and clones the two with an expire date <= current date.

Fixing that, the PR is approved

mi-kon:
@coldrunKacper please change the control title to  ‘Renewed document creation date offset (days)’

PR #94, opened 2016-08-29
PCHR-1527: Fix add to assignment input when adding documents

omarabuhussein:
## Problem 

When trying to search in  (add to assignment) box , it was not showing any result and a the following error appear in the console  :

```"Error in resource configuration for action `query`. Expected response to contain an array but got an object (Request: GET `/index.php?q=civicrm/case/ajax/unclosed)"```

The reason behind this is the deletion of this ajax end-point **civicrm/case/ajax/unclosed** in civicrm 4.7 and  a new quick-form method created **addEntityRef** to replace it . (https://github.com/civicrm/civicrm-core/commit/abd06efc426540ed16cb091093f90e13df2579f1)

## Solution

since we are working with angular and **addEntityRef** cannot be used since it is a quick-form method , I restored the **civicrm/case/ajax/unclosed** and the callback method **unclosedCases** in T&A extension.

So now the results are avaliable , but there is still a problem with the front-end part that need to be fixed in order for the fetched data to be shown.

selection_101

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Assignment.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Assignment.php
@@ -21,4 +21,83 @@ class CRM_Tasksassignments_BAO_Assignment extends CRM_Tasksassignments_DAO_Assig
 
     return $instance;
   } */
+
+
+  /**
+   * Retrieve cases along with related contact details
+   *
+   * @param array $params Accept :
+   *  - 'sort_name' : Sort name for the user you want to retrieve cases for.
+   *  - 'limit' : The number of cases you want to retrieve.
+   * @param array $excludeCaseIds
+   *
+   * @return array Of case and related data keyed on case id
+   */
+  public static function retrieveCases($params = array(), $excludeCaseIds = array()) {
+    if ($sortName = CRM_Utils_Array::value('sort_name', $params)) {
+      $config = CRM_Core_Config::singleton();
+      $search = ($config->includeWildCardInName) ? "%$sortName%" : "$sortName%";
+      $where[] = "( sort_name LIKE '$search' )";
+    }
+    if (is_array($excludeCaseIds) &&
+      !CRM_Utils_System::isNull($excludeCaseIds)
+    ) {
+      $where[] = ' ( ca.id NOT IN ( ' . implode(',', $excludeCaseIds) . ' ) ) ';
+    }
+
+    $where[] = ' ( ca.is_deleted = 0 OR ca.is_deleted IS NULL ) ';
+
+    //filter for permissioned cases.
+    $filterCases = array();
+    $doFilterCases = FALSE;
+    if (!CRM_Core_Permission::check('access all cases and activities')) {
+      $doFilterCases = TRUE;
+      $session = CRM_Core_Session::singleton();
+      $filterCases = CRM_Case_BAO_Case::getCases(FALSE, $session->get('userID'));
+    }
+    $whereClause = implode(' AND ', $where);
+
+    $limitClause = '';
+    if ($limit = CRM_Utils_Array::value('limit', $params)) {
+      $limitClause = "LIMIT 0, $limit";
+    }
+
+    $query = "
+    SELECT  c.id as contact_id,
+            c.sort_name,
+            ca.id,
+            ca.subject as case_subject,
+            civicrm_case_type.title as case_type,
+            ca.start_date as start_date,
+            ca.end_date as end_date,
+            ca.status_id
+      FROM  civicrm_case ca INNER JOIN civicrm_case_contact cc ON ca.id=cc.case_id
+ INNER JOIN  civicrm_contact c ON cc.contact_id=c.id
+ INNER JOIN  civicrm_case_type ON ca.case_type_id = civicrm_case_type.id
+     WHERE  {$whereClause}
+  ORDER BY  c.sort_name, ca.end_date
+            {$limitClause}
+";
+    $dao = CRM_Core_DAO::executeQuery($query);
+    $statuses = CRM_Case_BAO_Case::buildOptions('status_id', 'create');
+
+    $cases = array();
+    while ($dao->fetch()) {
+      if ($doFilterCases && !array_key_exists($dao->id, $filterCases)) {
+        continue;
+      }
+      $cases[$dao->id] = array(
+        'sort_name' => $dao->sort_name,
+        'case_type' => $dao->case_type,
+        'contact_id' => $dao->contact_id,
+        'start_date' => $dao->start_date,
+        'end_date' => $dao->end_date,
+        'case_subject' => $dao->case_subject,
+        'case_status' => $statuses[$dao->status_id],
+      );
+    }
+    $dao->free();
+
+    return $cases;
+  }
 }
--- a/uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/AssignmentTest.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/AssignmentTest.php
@@ -1,0 +1,164 @@
+<?php
+
+use Civi\Test\HeadlessInterface;
+use Civi\Test\TransactionalInterface;
+
+/**
+ *
+ * @group headless
+ */
+class CRM_Tasksassignments_BAO_AssignmentTest extends \PHPUnit_Framework_TestCase implements HeadlessInterface, TransactionalInterface {
+
+  use TasksassignmentsTestTrait;
+
+  private $contact_id;
+
+  private $case_type_id;
+
+  private $open_status_id;
+
+  public function setUpHeadless() {
+    return \Civi\Test::headless()
+      ->installMe(__DIR__)
+      ->apply();
+  }
+
+  public function setUp() {
+    // Create Contact
+    $params = ['first_name' => 'micky', 'last_name' => 'mouse'];
+    $this->contact_id = $this->createContact($params);
+    // Create Case Type
+    $params = ['name' => 'test_case_type', 'title' => 'Test Case Type', 'is_active' => 1];
+    $this->case_type_id = $this->createCaseType($params);
+    // Create Case Status option values
+    $params = ['name' => 'Opened', 'grouping' => 'Opened'];
+    $this->open_status_id = $this->createOptionValue($params, 'case_status');
+  }
+
+  public function testRetrieveCasesWithoutExcludeCaseIds() {
+    $casesFixtures = [
+      ['subject' => 'test case', 'end_date' => null],
+      ['subject' => 'test case2', 'end_date' => date('Ymd')]
+    ];
+
+    $defaultParams = [
+      'creator_id' => $this->contact_id,
+      'case_type_id' => $this->case_type_id,
+      'status_id' => $this->open_status_id,
+    ];
+
+    $createdCases = [];
+    foreach($casesFixtures as $caseFixture) {
+      $params = array_merge($defaultParams, $caseFixture);
+      $caseID = $this->createCase($params);
+      $createdCases[$caseID] = $params;
+    }
+
+    $cases = CRM_Tasksassignments_BAO_Assignment::retrieveCases();
+
+    $this->assertCount(count($createdCases), $cases);
+    foreach($cases as $caseID => $caseValues) {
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Assignment.php [hunk 1]
```diff
@@ -21,4 +21,91 @@ public static function create($params) {
...
+  /**
+   * Get all cases with no end dates
+   *
+   * @param array $params
+   * @param array $excludeCaseIds
+   * @param bool $excludeDeleted
+   *
+   * @return array of case and related data keyed on case id
+   */
+  public static function getUnclosedCases($params = array(), $excludeCaseIds = array(), $excludeDeleted = TRUE, $includeClosed = FALSE) {
```
davialexandre: I think that, instead of just restoring the deleted code, we should take the this as an opportunity to improve things. Searching for the usage of the civicrm/case/ajax/unclosed endpoint, I could see it's only used by the AssignmentSearch service. This service is then used by the search method of the AssignmentService, which only accepts one filter (besides the search term) to of IDs to be excluded. With all that given, I believe this method should be simplified to reflect this and we should drop $excludeDeleted and $includeClosed. 
The $params parameter could also be simplified to only accept the sort_name param, which is the only one used by the AssignmentService.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Assignment.php
```diff
@@ -21,4 +21,91 @@ public static function create($params) {
...
+            ca.start_date as start_date,
+            ca.end_date as end_date,
+            ca.status_id
+      FROM  civicrm_case ca INNER JOIN civicrm_case_contact cc ON ca.id=cc.case_id
+ INNER JOIN  civicrm_contact c ON cc.contact_id=c.id
+ INNER JOIN  civicrm_case_type ON ca.case_type_id = civicrm_case_type.id
+     WHERE  {$whereClause}
+  ORDER BY  c.sort_name, ca.end_date
+            {$limitClause}
+";
```
davialexandre: By reducing the responsibilities of method, maybe we can use the API here instead of depending on the internal cases structure? We would just have to make sure that, in doing that, the permissions will be applied correctly. That is, user X will only see her own tasks (or the tasks she has access permissions)

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/AssignmentTest.php
```diff
@@ -0,0 +1,81 @@
...
+  private $closed_status_id;
+
+  public function setUpHeadless() {
+    return \Civi\Test::headless()
+      ->installMe(__DIR__)
+      ->apply();
+  }
+
+  public function setUp() {
+    parent::setUp();
```
davialexandre: No need for this since your extending PHPUnit_Framework_TestCase, which has an empty setUp implementation.

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/AssignmentTest.php
```diff
@@ -0,0 +1,81 @@
...
+    $params = ['name' => 'test_case_type', 'title' => 'Test Case Type', 'is_active' => 1];
+    $this->case_type_id = $this->createCaseType($params);
+    // Create Case Status option values
+    $params = ['name' => 'Opened', 'grouping' => 'Opened'];
+    $this->open_status_id = $this->createOptionValue($params, 'case_status');
+  }
+
+  public function tearDown() {
+    parent::tearDown();
+  }
```
davialexandre: No need for this as well, for the same reason of setUp method

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/TasksassignmentsTestTrait.php
```diff
@@ -0,0 +1,82 @@
...
+    return NULL;
+  }
+
+  /**
+   * Creates single (Individuals) contact from the provided data.
+   *
+   * @param array $params should contain first_name and last_name
+   * @return int return the contact ID
+   * @throws \CiviCRM_API3_Exception
+   */
```
davialexandre: This doc block doesn't match what the method is doing

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/AssignmentTest.php
```diff
@@ -0,0 +1,81 @@
...
+      'end_date' => date('Ymd')
+    ];
+    $this->createCase($params);
+
+    $params = [
+      'case_type' => 'Test Case Type',
+    ];
+    $cases = CRM_Tasksassignments_BAO_Assignment::getUnclosedCases($params, [], TRUE, $includeClosed);
+
+    $this->assertCount($expectedCount, $cases);
```
davialexandre: This is not a good test. You're only asserting that you have the expected amount, but not the correct cases. If getUnclosedCases returns an array like [1, 2] it will still pass. 

Also, If you decide not to simplify getUnclosedCases, you need to add tests to cover all the possible combinations of the filters (one more reason to simplify it! :) )

uk.co.compucorp.civicrm.tasksassignments/xml/Menu/tasksassignments.xml
```diff
@@ -87,4 +87,10 @@
     <title>FileZip</title>
     <access_arguments>access Tasks and Assignments</access_arguments>
   </item>
+
+  <!-- API endpoint for angular: unclosed cases -->
```
davialexandre: I think this is an Ajax endpoint, not an API. But, anyway, this comment is kind of useless. What if I want to use for something not angular related?

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Assignment.php
omarabuhussein: yes I agree

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Assignment.php
omarabuhussein: the thing is what is happening in this query is not possible to do with single API call such as retrieving contact sort name , so using the API will reduce the performance of this method .

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/TasksassignmentsTestTrait.php
omarabuhussein: oops , I forget to fix TasksassignmentsTestTrait method docblocks

uk.co.compucorp.civicrm.tasksassignments/xml/Menu/tasksassignments.xml
omarabuhussein: yeah my fault :D  , I just copied it with paying much attention .

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Assignment.php
davialexandre: hmmmmm... I was thinking that maybe you could use the JOIN options on the API to fetch the data. Something like this:
captura de tela 2016-08-30 as 07 43 30

but I just tried it here and it's giving me an error (maybe a bug in civicrm or me doing something wrong):
captura de tela 2016-08-30 as 07 45 09

Anyway, if it's really not possible to do it using the API or if it will hurt performance too much, then I think it's fine to keep using a SQL query

omarabuhussein: naah , civicrm is buggy and unreliable :D ... the same problem with me when you set (case_id not in ..etc ) parameter ... also it appear that the (Like) clause don't work very well .

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Assignment.php
```diff
@@ -24,36 +24,26 @@ public static function create($params) {
 
 
   /**
-   * Get all cases with no end dates
+   * retrieve cases along with related contact details
```
davialexandre: Please start the sentence with an uppercase letter

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/AssignmentTest.php
```diff
@@ -17,42 +17,57 @@ class CRM_Tasksassignments_BAO_AssignmentTest extends \PHPUnit_Framework_TestCas
...
+      'subject' => 'test case2',
+      'status_id' => $this->open_status_id,
+      'end_date' => date('Ymd')
+    ];
+    $testCaseID2 = $this->createCase($params);
+
+    $cases = CRM_Tasksassignments_BAO_Assignment::retrieveCases();
+
+    $this->assertCount(2, $cases);
+    $this->assertTrue(!empty($cases[$testCaseID1]));
```
davialexandre: You can use the assertNotEmpty here `$this->assertNotEmpty($cases[$testCaseID1]);`

Even though this new assertion improves the test, I'm afraid this is still not good enough. If retrieveCases returns an array like
```
[
 1 => [1],
 2 => [2]
]
```

the test will still pass. Here's a suggestion of how to improve this test (also applies for the others):
```php
 $casesFixtures = [
  ['subject' => 'test case', 'end_date' => null],
  ['subject' => 'test case2', 'end_date' => date('Ymd')]
];

$defaultParams = [
  'creator_id' => $this->contact_id,
  'case_type_id' => $this->case_type_id,
  'status_id' => $this->open_status_id,
];

$createdCases = [];
foreach($casesFixtures as $caseFixture) {
  $params = array_merge($defaultParams, $casesFixtures);
  $caseID = $this->createCase($params);
  $createdCases[$caseID] = $params;
}

$cases = CRM_Tasksassignments_BAO_Assignment::retrieveCases();
$this->assertCount(count($createdCases), $cases);
foreach($cases as $caseID => $caseValues) {
  $this->assertNotEmpty($createdCases[$caseID]);
  $this->assertEquals($createdCases[$caseID]['subject'], $caseValues['case_subject']);
  $this->assertEquals($this->contact_id, $caseValues['contact_id']);
  $this->assertEquals($this->open_status_id, $caseValues['status_id']);
  // any other assertion that you might find relevant
}
```

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/TasksassignmentsTestTrait.php
```diff
@@ -23,23 +24,27 @@ protected function createContact($params) {
   }
 
   /**
-   * Creates single (Individuals) contact from the provided data.
+   * Create case type
    *
-   * @param array $params should contain first_name and last_name
-   * @return int return the contact ID
-   * @throws \CiviCRM_API3_Exception
+   * @param array $params should contain 'name'
```
davialexandre: Please start the $param description with an upper case letter in order to make it easier to spot where it starts: Should contain 'name'. The same applies to the doc blocks on the other methods

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Assignment.php
```diff
@@ -24,36 +24,26 @@ public static function create($params) {
...
    *
-   * @param array $params
+   * @param array $params accept 'sort_name' and 'limit'
```
davialexandre: I'm not sure here, but I'm thinking that, instead of this $params parameter, it would be better to have two separate parameters. What do you think.

In any case, it would be good to better describe what those $params are. It's easy to get what limit does, but sort_name is not that self-descriptive.

omarabuhussein: maybe in the future a new parameters will be added and used for othr sorts of  filtering , so I would keep it like than rather than keep adding new parameters to the method itself.

davialexandre:
Approved 👍

PR #95, opened 2016-08-30
PCHR-1360: Filters modified for documents

deb1990:
**Requirement**: Filters needs to be changed for Documents

**Fix**:
1. Common filter code were written in sidebar-filters.js which was getting shared by both task and documents.
2. Moved filter code to respective controllers and changed the filter for documents.
3. Test cases updated

image


kkalamarski:
Approved

PR #97, opened 2016-09-02
PCHR-1322: Removing "set status" option from document status dropdown

danilo-compucorp:
This PR removes the 'set status' option from document status dropdown and uses the "_ui-select_" dropdown, with the "_allow-clear_" option.

Before:
before

After:
pchr-1322

This PR is related to: https://github.com/civicrm/civihr/pull/1130


brunoqueiros:
Approved

PR #102, opened 2016-09-09
PCHR-1523: form validations

danilo-compucorp:
This PR updates the form validation styles, making mandatory fields be indicated with red asterisk, and adding red border for missing required fields.
It also changes the style of the standard notifications.

### Add task

**before**
add-task-before

**after**
add-task

### Add assignment

**before**
add-assigment-before

**after**
add-assigment

### Add document

**before**
add-document-before

**after**
add-document

### Alert

**before**
alert-before

**after**
alert

### Success

**before**
success-before

**after**
success

### Info

**before**
info-before

**after**
info

### Error

**before**
error-before

**after**
error

This PR is related to https://github.com/civicrm/civihr/pull/1150


brunoqueiros:
Approved

JonScreat:
- Colour of asterisks should be #E6807F
- Date picker - end of the date is missing when the field is red/required
- Asterisks shouldn’t show after failed validation if it’s not there before. Asterisks are used to show that a field is required - the red outlines are used to show that the field has been missed - so if there isn’t an asterisk there before there shouldn’t be one after.

danilo-compucorp:
@JonScreat the asterisks are always shown for required fields, both before and after validation. 
Not sure what you mean by:

> end of the date is missing when the field is red/required

JonScreat:
On some of the date fields the year is missing.
image

Understood regarding the asterisks. I think this might need some further thought from a UX point if view as it's a little overbearing and I'm not sure how well this works for the fields without labels - I wikk discuss this with Mike.

Also would it be possible to move this asterisk?
image

JonScreat:
Hi Danilo,

We've had a discussion around this and updated the validation rules:

Mandatory fields should be indicated with red asterisk:

- If field has a label - show asterisk after the label
- If field doesn't have a label - shod asterisk after the placeholder text
- If there is a set of fields in a table - show asterisk after the table column header (not each field individually).

1. If user tries to submit the form with blank mandatory fields - user red border for missing fields. If fields are on another tab - highlight tab title
2. Save button should always be enabled
3. If field requirements are not met (e.g. email format is incorrect) - show an error message in label label-danger format
4. If form cannot be saved due to failed server validation - display error message using new notification design, see https://compucorp.atlassian.net/wiki/display/PCHR/Consistent+approach+to+notifications+needed?moved=true

PR #104, opened 2016-09-13
PCHR-1568: Fix fading effect on tab switch

AkA84:
## Problem
When switching tabs in T&A, you can momentarily see the old tab still present at the top of the page even when the new tab has already been loaded

before

This is due to a custom `.fade-in-up` class w/ animation that conflicts with the DOM manipulation that angular performs while switching view, basically forcing that view to still be visible before being removed

## Solution
Making sure that the `.fade-in-up` (and similar `.fade-in`) classes are not applied to elements that are about to be removed by angular (`.ng-leave`) does the trick

after

## Problem
There were a couple of tests failings for the `ModalDocumentCtrl` controller, related to `$scope.statusFieldVisible` and `$scope.showStatusField`.

Turns out that the implementation of this properties was changed by #97 and the tests weren't updated.

After further inspection to the code it seemed to me that there was actual no use at all anymore for neither of them, given that the "status" field will be always present anyway, and will always have a value thanks to this line:
```js
$scope.document.status_id = $scope.document.status_id || '1';
```

## Solution
I just remove everything related to those two properties


danilo-compucorp:
Approved

PR #113, opened 2016-10-05
Sync master with integration

AkA84:
@coldrunKacper 
**PCHR-1365** Create new document before pre-set days before original expiry date

@AkA84
**PCHR-1568** T&A tabs appear  one under another when switching between them 
**PCHR-1584** Load extensions resources only when their tab is loaded

@danilo-compucorp 
**PCHR-1523** Update validation behaviour in popup forms 
**PCHR-1524** datepicker update


AkA84:
signing off on behalf of danilo

PR #136, opened 2017-02-01
PCHR-1904: Changing activity URL depending upon contact id permissions.

hitesh-jain:
**Problem:**
The daily email digest goes out to all creators (civihr_admin) and assignees (could be civihr_admin/civihr_manager/civihr_staff) of tasks/docs that are overdue/due today/coming up this week. Creators see both tasks/docs created and delegated by them. Assignees see only tasks/docs they are supposed to work on.
Need to use different destination urls depending on user having access civicrm permission.

**Solution:**
Check the contact id loading corresponding drupal user for access permission "access CiviCRM" , if it has access than the email link should point to civicrm else it should link to dashboard.

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
@@ -409,7 +409,7 @@ class CRM_Tasksassignments_Reminder {
         if ($reminderKey) {
           $reminderData[$reminderKey][] = array(
             'id' => $activityResult->id,
-            'activityUrl' => CIVICRM_UF_BASEURL . '/civicrm/activity/view?action=view&reset=1&id=' . $activityResult->id . '&cid=' . $contactId . '&context=activity&searchContext=activity',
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
```diff
@@ -478,6 +478,30 @@ public static function canSeeKeyDates($contactId) {
...
+   * @return string
+   */
+  public static function createActivityURL($contactId = NULL) {
+
+    if ($contactId) {
+      $activityUrl = '';
+      $drupal_user = user_load(_get_uf_match_contact($contactId)['uf_id']);
+
+      if (user_access('access CiviCRM', $drupal_user)) {
+        $activityUrl = CIVICRM_UF_BASEURL . '/civicrm/activity/view?action=view&reset=1&id=' . $activityResult->id . '&cid=' . $contactId . '&context=activity&searchContext=activity';
```
coldrunKacper: If you want to use $activityResult here then you need to pass it into the function on line 412.

coldrunKacper:
Approved.

PR #138, opened 2017-02-03
PCHR-1871: Redirect to Tasks Dashboard on Page not Found

MiyaNoctem:
## Problem
Accessing non existing pages in civicrm application results in showing invalid data, being showed by default CiviCRM dashboard.

before

## Solution
Fixed by overriding default "civicrm" path to use CiviHR Tasks & Assignments dashboard instead.

after


uk.co.compucorp.civicrm.tasksassignments/xml/Menu/tasksassignments.xml
```diff
@@ -1,5 +1,17 @@
...
+    <path>civicrm</path>
+    <title>CiviCRM</title>
+    <access_callback>CRM_Core_Permission::checkMenu</access_callback>
+    <access_arguments>access CiviCRM</access_arguments>
+    <page_callback>CRM_Tasksassignments_Page_Dashboard</page_callback>
+    <page_arguments>null</page_arguments>
+    <is_ssl>false</is_ssl>
+    <weight>0</weight>
+  </item>
+
```
davialexandre: The look looks good, but I'm not sure if this is the best approach for this. My concerns are:
- This is not a real redirect, but you're "hijacking" the url to return what you want
- If, for some reason, The Page_Dashboard does the assumption it will always be under civicrm/tasksandassignments/dashboard, it will break when accessed from a non-existent url
- As far as I know, the T&A extension is supposed to work as standalone extension. That is, if I have an existing civicrm site I can just install T&A without all the other things from CiviHR. Those users would probably be mad to know that they cannot see they civicrm dashboard anymore.

My opinion is that, if a page doesn't exist we should tell the user that without any redirects. This is a web application and it's how HTTP works. For this reason, I think an ideal solution would be to think about the possibility of adding a new hook to civicrm which would be triggered when the user hits a non-existent page. Or maybe we could use one of the hooks available?

Independent of the solution, I think the redirect (a real redirect) should be done only for civihr. So I think that the proposed solution on the ticket is better than this one. What's the reason for changing it for this one?

MiyaNoctem:
Going to implement as part of CiviHR core so that T&A stand-alone users can keep working with CiviCRM default dashboard.

PR #144, opened 2017-02-15
PCHR-1872: Fix Error on T&A Calendar View

MiyaNoctem:
## Problem
Expected behaviour for 'get' API call to Assignments entity in front-end required it to return an array of contacts for each assignment found, however, this was not happening. As it turned out, Assignments entity is based on CiviCRM Case entity, and this contact array is only included in results if a case ID is passed as parameter on API call.  However, CiviCRM wasn't able to identify the ID field for the Assignments entity appropriately, expecting it to be either 'id' or 'assignment_id', which never happened, since it was actually 'case_id', as Assignments DAO was a direct extension of Case BAO.

image-2017-01-12-16-42-09-824

## Solution
Fixed the problem by creating 'id' field in Assignment's DAO, by altering the fields array defined in Case DAO and copying values for 'case_id' and changing its key to 'id'.

after


coldrunKacper:
Approved.

PR #145, opened 2017-02-27
PCHR-1888: Remove Deleted Contacts' Info From Key Dates

MiyaNoctem:
# Problem 
Contacts deleted to trash still appear on key dates. They should not. 

# Solution
API Call to KeyDates Entity 'get' function returned dates associated to contacts in trash can, since the is_deleted field was not taken into account when building the query.  Fixed by adding 'is_deleted = 0' to all queries used to obtain key dates.

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
@@ -1,0 +1,43 @@
+<?php
+use Civi\Test\HeadlessInterface;
+use Civi\Test\TransactionalInterface;
+use CRM_HRCore_Test_Fabricator_Contact as ContactFabricator;
+
+/**
+ * Tests for CRM\Tasksassignments\KeyDates Class of Tasks & Assignments extension
+ * 
+ * @group headless
+ */
+class CRM_Tasksassignments_KeyDatesTest extends PHPUnit_Framework_TestCase implements
+  HeadlessInterface,
+  TransactionalInterface {
+  
+  public function setUpHeadless() {
+    return \Civi\Test::headless()
+      ->install('uk.co.compucorp.civicrm.hrcore')
+      ->installMe(__DIR__)
+      ->apply();
+  }
+  
+  public function testDatesForDeletedContactsNotInResult() {
+    $contact = ContactFabricator::fabricate(['birth_date' => '1977-06-27']);
+    ContactFabricator::fabricate(['birth_date' => '1978-06-19']);
+    ContactFabricator::fabricate(['birth_date' => '1982-08-16']);
+
+    $query = '
+      UPDATE civicrm_contact
+      SET is_deleted = 1
+      WHERE id = %1
+    ';
+    $params = [1 => [$contact['id'], 'Int']];
+    CRM_Core_DAO::executeQuery($query, $params);
+
+    $keyDates = civicrm_api3('KeyDates', 'get', [
+      'sequential' => 1,
+    ]);
+
+    foreach ($keyDates['values'] as $currentDate) {
+      $this->assertNotEquals($contact['id'], $currentDate['contact_id']);
+    }
+  }
+}
--- a/uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
@@ -20,24 +20,19 @@ class CRM_Tasksassignments_KeyDatesTest extends PHPUnit_Framework_TestCase imple
   }
   
   public function testDatesForDeletedContactsNotInResult() {
-    $contact = ContactFabricator::fabricate(['birth_date' => '1977-06-27']);
-    ContactFabricator::fabricate(['birth_date' => '1978-06-19']);
-    ContactFabricator::fabricate(['birth_date' => '1982-08-16']);
-
-    $query = '
-      UPDATE civicrm_contact
-      SET is_deleted = 1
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
```diff
@@ -0,0 +1,43 @@
...
+  
+  public function setUpHeadless() {
+    return \Civi\Test::headless()
+      ->install('uk.co.compucorp.civicrm.hrcore')
+      ->installMe(__DIR__)
+      ->apply();
+  }
+  
+  public function testDatesForDeletedContactsNotInResult() {
+    $contact = ContactFabricator::fabricate(['birth_date' => '1977-06-27']);
```
tunbola: Won't it be simpler to just do this `ContactFabricator::fabricate(['birth_date' => '1977-06-27', 'is_deleted' => 1])` ?

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
```diff
@@ -0,0 +1,43 @@
...
+    ContactFabricator::fabricate(['birth_date' => '1978-06-19']);
+    ContactFabricator::fabricate(['birth_date' => '1982-08-16']);
+
+    $query = '
+      UPDATE civicrm_contact
+      SET is_deleted = 1
+      WHERE id = %1
+    ';
+    $params = [1 => [$contact['id'], 'Int']];
+    CRM_Core_DAO::executeQuery($query, $params);
```
tunbola: We can do away with this query if the Contact is fabricated as deleted right from the onset.

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
```diff
@@ -0,0 +1,43 @@
...
+
+    $keyDates = civicrm_api3('KeyDates', 'get', [
+      'sequential' => 1,
+    ]);
+
+    foreach ($keyDates['values'] as $currentDate) {
+      $this->assertNotEquals($contact['id'], $currentDate['contact_id']);
```
tunbola: We should validate here that the API count returns two and also that the contactID's are those of the two contacts who were not deleted.

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
```diff
@@ -20,24 +20,19 @@ public function setUpHeadless() {
...
     $keyDates = civicrm_api3('KeyDates', 'get', [
       'sequential' => 1,
     ]);
 
+    $this->assertEquals(sizeof($contacts), $keyDates['count']);
+
+    // Verify deleted contact is not in result
     foreach ($keyDates['values'] as $currentDate) {
-      $this->assertNotEquals($contact['id'], $currentDate['contact_id']);
+      $this->assertContains($currentDate['contact_id'], $contacts);
```
tunbola: I think it will actually be better to use assertEquals here so as to be more specific.
the contacts array will be something like this `[0 => ['id'=>1, 'name'=>'blabkla'], 1 =>['id' => 2, 'name' =>'blabla'] ]`  and we are validating that it contains the contact_id.  To be safer we can do two assertions
$this->assertEquals($keyDates['values'][0]['contact_id'], $contacts[0]['contact_id'])
$this->assertEquals($keyDates['values'][1]['contact_id'], $contacts[1]['contact_id'])` depending on the order the KeyDates API returns the results.

MiyaNoctem: Currently, keydates are returned ordered by ascending date, but I don't think it's a good idea to have the test depend on this order.  If for any reason this ordered is changed in the future, the test would fail, when it shouldn't.

tunbola: We can actually force the API to return in the order we want using the sort parameter.
Not just comfortable using assertContains on an array of arrays, its more prone to mistakes. It's much easier when its a single value against a simple array.

tunbola: Another way is we can keep the loop and do something like this.
`$contactIDs = [$contacts[0]['contact_id'], $contacts[1]['contact_id'])
foreach ($keyDates['values'] as $currentDate) {
$this->assertContains($currentDate['contact_id'], $contactIDs);
}`
So that will keep it simple.

tunbola: We can use assertNotContains with the loop too since we are more concerned about the deleted contact.

MiyaNoctem: It's not an array of arrays, it's an array of integers.  Check the '['id']' at the end of the lines:
```
$contacts[] = ContactFabricator::fabricate(['birth_date' => '1978-06-19'])['id'];
$contacts[] = ContactFabricator::fabricate(['birth_date' => '1982-08-16'])['id'];
```
Still, I understand there's a readibility problem there.  Ill make it cleaner.

tunbola: Sure. And I actually didn't notice the id appended at the end. :).

PR #147, opened 2017-02-28
PCHR-1923: Fix assignment modal

AkA84:
# Problem
The bulk assignment of assignees and due dates always try to use the first task data, but if the first task is disabled, then an error occurs

bulk-assignment-before

# Solution
The bulk assignment now uses the first *enabled* task. Additionally, if the first enabled task doesn't have the assignee or due date assigned, then the bulk assignment will deselect assignee/due date also for the rest of the tasks

bulk-assignment-after

---

# Problem
When changing the key date of the assignment, the due date of the tasks will change accordingly. The problem is that the due date will be assigned also to the disabled tasks

key-date-watch-before

# Solution
The disabled tasks will now be ignored when the assignment key date is changed

key-date-watch-after

---

# Problem
When a task was disabled, an error will occur during the validation process, which would result in a JS error

save-with-disabled-before

# Solution
The validation code now ignores the disabled tasks

save-with-disabled-after

---

# Problem
When saving an assignment, the due dates of the assignment's tasks/documents appeared to be cleared, triggering a "required" style on the due dates fields

due-dates-save-before

# Solution
The issue was caused by the fact that the task/document objects were passed to the `TaskService`/`DocumentService`, which would amend the date format for storing it in the DB. Given that the change on the date field was done on the same objects that were also displayed on the view, this caused the dates to disappear (as the datepicker would not be able to display the date in the server's date format)

Passing a copy of the objects to the services, instead of the objects themselves, solved the issue

due-dates-save-after


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
```diff
@@ -148,136 +149,137 @@ define([
             };
 
             $scope.confirm = function () {
+              if (!($filter('filter')($scope.taskList, { create: true })).length && !($filter('filter')($scope.documentList, { create: true })).length) {
```
deb1990: Please give a line break, preferably after `&&`

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
```diff
@@ -342,31 +361,33 @@ define([
                 missingRequiredFields.push('Key date');
               }
 
-              if ($scope.taskList.filter(function(activity){
-                return !activity.activity_type_id;
+              if (enabledTasks.filter(function (task) {
```
deb1990: We are filtering 3 times that means 3 loops, we can optimize this code using a single loop,
```javascript
_.each(enabledTasks, function(task){
 if(!task.activity_type_id && missingRequiredFields.indexOf('Activity type') === -1){ 
   missingRequiredFields.push('Activity type')
 }
 //Same for Assignee and Activity Due Date
})
```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
```diff
@@ -148,136 +149,137 @@ define([
...
-                    return;
-                }
+              $scope.$broadcast('ct-spinner-show');
+              $scope.assignment.start_date = new Date();
 
-                if (!validateRequiredFields($scope.assignment)) {
-                  return;
-                }
+              AssignmentService.save($scope.assignment).then(function (resultAssignment) {
+                var documentListAssignment = $scope.documentList.filter(function (doc) {
```
deb1990: To remove  loop looping twice, same for `taskList`
```javascript
var documentListAssignment = [];
_.each($scope.documentList.filter, function(){
 if(doc.create) {
  doc.case_id = resultAssignment.id;
  documentListAssignment.push(doc);
 }
})
```

PR #152, opened 2017-03-07
PCHR-1925: Fixed Additional documnent or task break new assignment screen UI

mukeshcompucorp:
## Problem
Additional document or task break new assignment screen UI

pchr-1925

### Fixed by #167  
Change markup for the structure to fix up the UI

resolved-pchr-1925

### Fixed By this PR:
Restructure Html markup so that the form does not go out of the modal. 
On selecting larger value for Task type in newly added task, the task type select option field remains with constant width.
constant-width-ui-select


uk.co.compucorp.civicrm.tasksassignments/views/modal/assignment.html
```diff
@@ -128,18 +129,20 @@ <h4 class="modal-title">New Assignment</h4>
...
-                        <span class="input-group-btn" ng-if="activity.isAdded">
-                          <a class="btn btn-default btn-danger" ng-click="removeActivity(taskList, $index)">
-                            <span class="fa fa-remove" aria-hidden="true"></span>
-                          </a>
-                        </span>
                     </div>
                   </td>
+                  <td align="center">
+                    <span class="input-group-btn" ng-if="activity.isAdded">
+                      <a class="btn btn-default btn-danger" ng-click="removeActivity(taskList, $index)">
```
deb1990: Are you sure that you need both `btn-default` and `btn-danger`?

uk.co.compucorp.civicrm.tasksassignments/views/modal/assignment.html
```diff
@@ -128,18 +129,20 @@ <h4 class="modal-title">New Assignment</h4>
...
                         <span class="input-group-addon fa fa-calendar"></span>
-                        <span class="input-group-btn" ng-if="activity.isAdded">
-                          <a class="btn btn-default btn-danger" ng-click="removeActivity(taskList, $index)">
-                            <span class="fa fa-remove" aria-hidden="true"></span>
-                          </a>
-                        </span>
                     </div>
                   </td>
+                  <td align="center">
+                    <span class="input-group-btn" ng-if="activity.isAdded">
```
deb1990: Why do you need this `input-group-btn` class for a simple button?
Also it seems that the span can be removed and the `a` can be directly placed inside `td`

uk.co.compucorp.civicrm.tasksassignments/views/modal/assignment.html
```diff
@@ -128,18 +129,20 @@ <h4 class="modal-title">New Assignment</h4>
...
                           ng-click="!isDisabled && activity.create ? dpOpen($event, 'task'+$index) : return;" ng-disabled="isDisabled || !activity.create">
                         <span class="input-group-addon fa fa-calendar"></span>
-                        <span class="input-group-btn" ng-if="activity.isAdded">
-                          <a class="btn btn-default btn-danger" ng-click="removeActivity(taskList, $index)">
-                            <span class="fa fa-remove" aria-hidden="true"></span>
-                          </a>
-                        </span>
                     </div>
                   </td>
+                  <td align="center">
```
deb1990: Dont use non-css properties to style. Instead you can use bootstrap class to do the same.

uk.co.compucorp.civicrm.tasksassignments/views/modal/assignment.html
bmnepali: `btn-default` seems to add little shadow, which seems expected as this was the previously return code.

uk.co.compucorp.civicrm.tasksassignments/views/modal/assignment.html
bmnepali: Right! will do.

uk.co.compucorp.civicrm.tasksassignments/views/modal/assignment.html
bmnepali: Don't need `input-group-btn` as we are placing the button separately. Also will place `<a ...> ... </a>` directly in `td`.

PR #157, opened 2017-03-09
PCHR-2052: Fix Calendar for Users Without Assigned Tasks

MiyaNoctem:
## Problem
When a user had no assigned tasks, but was the source for some tasks, API calls to obtain task lists were returning an empty array for assigned tasks, while returning an object for list of tasks where current user was the source. This inconsistency caused problems in front-end, making it uncapable of dealing with an array vs an object for each call.

This happens because, although the back-end was returning an array on all calls, calls without the sequential flag generate an associative array, which is sent as an object in json, since there are no associative arrays in javascript.

## Solution
Fixed by sending sequential flag = 1 for all calls to get Tasks.  This way, the result will always be an array.


davialexandre:
@MiyaNoctem a tip for the next PRs: try to put refactoring in a separate commit. I was taking a look at this and, at first, I thought you kept the backend solution and added a frontend solution (basically, you had 2 solutions for the same problem). Only after comparing the files side-by-side I realized that the PHP changes were just a refactoring you did. 

Separate commits keep the changes more organized and help reviewers to understand what is going on.

PR #158, opened 2017-03-09
PCHR-1908: Fix due date field size in add task popup

mukeshcompucorp:
## Problem
Cutting date in due date field

ticket-pchr-1908

## Solution
Change css padding: 6px 12px; to padding: 6px 8px; in date field.

resolved-pchr-1908


AkA84:
@mukeshcompucorp alternative approach

screen

Instead of having to do an ad-hoc-with-magic-numbers css fix, you can simply shorten the label ("date" in "due date" to me is redundant, given that you have a datepicker right next to it so it is obvious what we're talking about), and tweak the `col-sm-*` bootstrap grid classes of both label and input. This way you don't have to override any default bootstrap style for such a specific case

AkA84:
@mukeshcompucorp also according to the ticket you should also do the same fix for the document modal. and in the document modal you should also fix the expiry date field. Something like
screen

mike already signed this off

PR #159, opened 2017-03-09
FAR-144: Remove the limit on internal call to api

mickadoo:
## Problem

In some cases creating new action types for an assignment led to the action types being disabled in the new assignment modal. 

The cause was that the call to the API to fetch all tasks for the CiviTask compontent was being limited to 25 results. In the frontend, validation for the activity types with missing tasks caused them to be disabled.

## Solution

- Add a parameter to remove the limit for the "get options" call from the frontend so it will return all options
- In the backend pass those parameters on internally in the "get options" endpoint so that a limit is still possible if desired

### Notes

There are also some whitespace changes since my IDE setting removes whitespace from empty lines.


uk.co.compucorp.civicrm.tasksassignments/api/v3/Task.php
```diff
@@ -611,8 +622,8 @@ function civicrm_api3_task_getstatuses($params) {
     return null;
 }
 
-function _civicrm_api3_task_gettypesbycomponent($component, $sequential = 1) {
-    
+function _civicrm_api3_task_gettypesbycomponent($component, $params = []) {
+
```
omarabuhussein: please remove this space

mickadoo: @omarabuhussein absolutely inconceivable, this space has to stay! :-) 

I wasn't sure whether to add a space since the other functions are a mix. Good to know for the future.

omarabuhussein:
@mickadoo please squash the last commit into the previous one

mickadoo:
@omarabuhussein commits squashed

PR #163, opened 2017-03-14
PCHR-1919: Hide action bar and checkboxes form document list

bmnepali:
### Issue:
1. Action Bar is present.
2. There are checkboxes in each document list and a checkbox to toggle between checkboxes in the the document list title.

### Solution:
1.  Remove Action bar related markup form `documents.html` and `calendar.documentList.html`
2. Remove checkboxes from list title and form every document in the list.
3. Remove related scope variables and functions  in controllers on removal of action bar and checkboxes.

#### Before:
1. Documents lists Page
screen shot 2017-03-14 at 2 12 34 pm

2. Document list page through calendar
screen shot 2017-03-14 at 5 28 09 pm

#### After:
1. Document List page
screen shot 2017-03-14 at 3 20 37 pm

2. Document list page through calendar
screen shot 2017-03-14 at 5 25 59 pm


AkA84:
@bmnepali you are removing parts of the functionality, shouldn't you also remove bits of controllers/services?

bmnepali:
@AkA84 Yes, I should remove the related codes in controller, but functions like `toggleAll()` and `toggleIsCheckedAll()`are used in T&A Dashboard -> Calendar Tab -> Documents list (in calendar.documentList.html) as well.
I just confirmed with Mike, that the action bar and checkboxes are to be removed form the list of documents through calendar tab i.e (form calendar.documentList.html). So now, when I remove the markup, I can safely remove codes form controllers well.

PR #164, opened 2017-03-14
PCHR-1913: Fix dropdown input focus

brunoqueiros:
# Problem
When you click any select dropdown in new task or new assignment or new document modal you cannot start typing search queries immediately, you need to click the field again.

# Solution
It was created a decorator to uiSelect directive to put a focus in the input when the dropdown is open.

**Before**

first-click

**After**

kapture 2017-03-14 at 7 06 56


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/directives/ui-select.js
```diff
@@ -0,0 +1,29 @@
...
+        $provide.decorator('uiSelectDirective', ['$delegate', '$timeout', function ($delegate, $timeout) {
+            var directive = $delegate[0];
+            var directiveCompile = directive.compile;
+
+            directive.compile = function (tElement, tAttrs) {
+              var link = directiveCompile.apply(this, arguments);
+
+              return function(scope, elem, attrs, uiSelect) {
+                  link.apply(this, arguments);
+                  scope.$watch('$select.open', function(val) {
```
bmnepali: Use `function (val) {..}` not `function(val) {...}`. As you have done above.

PR #165, opened 2017-03-15
PCHR-1956: Fix assignee field

brunoqueiros:
# Problem
If you go to tasks tab on contact page and go to assignee field. You can see there is one hidden field there and when you change assignee sometimes it's not displayed.

# Solution
Removed the unused field and update the contacts cache to show the AJAX selected contact.

**Before**

task-tab

**After**

kapture 2017-03-15 at 9 18 40


uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/tasks-list_test.js
```diff
@@ -45,6 +46,42 @@ define([
...
+            updateObj = {
+              subject: 'Test 2'
+            };
+
+            saveObj = angular.extend({}, task, updateObj);
+
+            $scope.updateTask(task, updateObj);
+          });
+
+          it('should call TaskService with the correct object', function () {
```
bmnepali: More preferable to state as **Calls TaskService with the correct object**

PR #168, opened 2017-03-16
PCHR-1920: Remove task assignee requirement

mickadoo:
## Problem

The assignee is currently a requirement when creating new tasks and documents. We want to allow people to create tasks and documents but decide the assignee later (or not assign anyone) so we will remove the assignee requirement

## Solution

Remove validation to check for assignee ID

## Notes

- This will also require frontend changes which will probably be done by @bmnepali. Without these changes the frontend validation will not allow this to be tested.


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -19,8 +19,8 @@ public static function create(&$params) {
 
     if ($hook === 'create') {
       // If creating a new Document, we require all three contacts to be defined.
```
omarabuhussein: this inline comment is not longer reflect the current behavior , a better one could be : 

// If creating a new Document, both target_contact_id and source_contact_id are required

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Task.php
```diff
@@ -13,8 +13,8 @@ public static function create(&$params) {
 
     if ($hook === 'create') {
         // If creating a new Task, we require all three contacts to be defined.
```
omarabuhussein: same goes here 

https://github.com/compucorp/civihr-tasks-assignments/pull/168/files#r106493070

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
```diff
@@ -131,14 +131,19 @@ private static function _extractPreviousAssignee(&$assignees, $previousAssigneeI
    * Gets the full list of the recipients of the reminder email
    * Makes sure there are no duplicates or no empty emails in the list
    *
-   * @param array $activityContacts
+   * @param array $contacts
    * @param array $previousAssignee
    * @return array
```
omarabuhussein: not related to your changes but can you add a space between @param and @return

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/DocumentTest.php
```diff
@@ -50,18 +50,6 @@ function testCreateDocumentWithNoSource() {
   /**
    * @expectedException CRM_Exception
    */
-  function testCreateDocumentWithNoAssignee() {
```
omarabuhussein: I think it is better to keep these tests but change them a a positive tests , so ensure that even if you created a document with no assignee it will still go through and the document will be created ... same thing go for the tests below .

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
```diff
@@ -131,14 +131,19 @@ private static function _extractPreviousAssignee(&$assignees, $previousAssigneeI
...
    */
-  private static function _reminderRecipients($activityContacts, $previousAssignee) {
-    return array_filter(array_unique(array_merge(
-                            $activityContacts['assignees']['emails'], $activityContacts['source']['emails'], [$previousAssignee['email']]
-    )));
+  private static function _reminderRecipients($contacts, $previousAssignee) {
+    $hasSources = !empty($contacts['source']['emails']);
```
omarabuhussein: a shorter way to write this would be :

```php
$assignees = $contacts['assignees']['emails'] ?: [];
$sources = $contacts['source']['emails'] ?: [];
$previous = !empty($previousAssignee['email']) ? [$previousAssignee['email']] : [];

$recipients = array_merge($assignees, $sources, $previous);

return array_filter(array_unique($recipients));
```

hence in your code that you forget to the 'email' index for $previous  variable , also you miss some spaces that make the code easier to read ... both corrected in my version

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/DocumentTest.php
mickadoo: I reverted the deletion and remove the `expectException` but I don't have testing set up locally so I can't be sure if these tests are passing now.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
mickadoo: Good catch with the missing `'email'` index

The `$assignees = $contacts['assignees']['emails'] ?: [];` is nice and short, but with errors on it can lead to an undefined notices if 'emails' is not set. I know it's not so terrible, but I would prefer not to have any notices in development because if we have 1000 of them maybe one will get through that's actually a bug.

mickadoo:
@omarabuhussein  Updated the comments, tests and left a reply regarding the method changes.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
omarabuhussein: ops , I though php is smart enough to not show a notice for such case but I was wrong .. I suggest to use what I've suggest but instead of using short ternary operator syntax use the long one and add !empty() to check if the value is their.

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/DocumentTest.php
omarabuhussein: removing expectException is not enough , you still need to do some assertion on the result  ..  I will run them on my machine once you fix this

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
mickadoo: `$assignees = !empty($contacts['assignees']['emails']) ?: [];` 

would lead to $assignees being `true`. 

`$assignees = !empty($contacts['assignees']['emails']) ? $contacts['assignees']['emails'] : [];`

Is not bad, but my main reason for moving the `empty` check into a different line was to keep the line length short and make it more readable.

omarabuhussein: yup I was talking about something like this :

```php
$assignees = !empty($contacts['assignees']['emails']) ? $contacts['assignees']['emails'] : [];
```

but I am fine with your version and keeping  line length less than 80 character as you mentioned on hipchat is good point actually .. but I think adding some spaces will make this a bit better :

```php
    $hasSources = !empty($contacts['source']['emails']);
    $hasAssignees = !empty($contacts['assignees']['emails']);

    $assignees = $hasAssignees ? $contacts['assignees']['emails'] : [];
    $sources = $hasSources ? $contacts['source']['emails'] : [];
    $previous = $previousAssignee ? [$previousAssignee] : [];

    $recipients = array_merge($assignees, $sources, $previous);
```

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/DocumentTest.php
mickadoo: Ah yes, silly of me. I added assertions - hopefully they work. I need to get testing set up on my machine soon

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
mickadoo: :+1: I added the vertical spacing.

PR #187, opened 2017-04-12
PCHR-2119: Only Allow Valid Assignment Tasks

mickadoo:
## Problem

You can add any kind of task to an assignment workflow, but these tasks will appear as disabled when you attempt to assign the assignment. In addition there is no validation on whether a task is deleted / inactive when adding it to an assignment type.

## Solution

- Hide the available task types in the assignment type edit screen to only CiviTask and CiviDocument
- Validate the input when saving an assignment type to allow only active CiviTask and CiviDocument types
- Hide inactive task types from the Assignment modal

## Before
pchr-2119-before

## After
pchr-2119-after


uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -12,6 +16,28 @@ function tasksassignments_civicrm_config(&$config) {
...
+  $isActivityType = $optionGroup === 'activity_type';
+  $entity = ArrayHelper::value('entity', $apiRequest);
+  $action = ArrayHelper::value('action', $apiRequest);
+
+  $jsonFromCaseEdit = '"actTypes":["OptionValue","get",{"option_group_id":"activity_type","options":{"sort":"name","limit":0}}]';
+  $isAssignmentEdit = strpos($requestJson, $jsonFromCaseEdit) !== FALSE;
+
+  if ($isAssignmentEdit && $entity === 'OptionValue' && $action === 'get' && $isActivityType) {
+    $wrappers[] = new StripInvalidTaskTypesWrapper();
+  } elseif ($entity === 'CaseType' && $action === 'create') {
```
omarabuhussein: since you only show activity types that belongs to civitask and cividoc components , I think there is no need to  do this check when creating caseType .. so I think you can remove this condition and the whole **AssignmentDefinitionValidationWrapper** class .

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -12,6 +16,28 @@ function tasksassignments_civicrm_config(&$config) {
 }
 
 /**
+ * @param array $wrappers
```
mickadoo: You mean to use "Implementation of.." instead of "Implements.."?

If I use that I get warnings from civilint:


> Format should be "* Implements hook_foo().", "* Implements hook_foo_BAR_ID_bar() for xyz_bar().",, "* Implements hook_foo_BAR_ID_bar() for xyz-bar.html.twig.", or "* Implements hook_foo_BAR_ID_bar() for xyz-bar.tpl.php.".

omarabuhussein: then use what civilint suggests :)

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
mickadoo: I would say that just hiding it is not enough. If we don't want to allow it then we should have validation to stop it. 

Also for some people who have already created assignments we should tell them if they already have activity types that are no valid.

omarabuhussein: > I would say that just hiding it is not enough. If we don't want to allow it then we should have validation to stop it.

I asked for that because the way in which you prevent creating it is a but dangerous , it apply to 'CaseType' create actions .. but hrrequirement extensions has some usages for CaseType entity  so I don't want to risk that until the relationship between them is more clear.

> Also for some people who have already created assignments we should tell them if they already have activity types that are no valid.

I don't think this a big problem now .. applying what you did to 'create' CaseType entity worrying me more .

mickadoo: I wasn't aware of the hrrequirement CaseType usages. If CaseType is being used elsewhere in CiviHR then this changes how we should handle this for sure.

I think we definitely should have strict validation because relying on frontend validation is just weak, and can easily be bypassed. If we have business logic that requires that users are not allowed to create assignment types with anything other than "CiviTask" or "CiviDocument" then I think we should enforce that on the backend.

This is another example of the problem with hacking the case page to our use for assignments, we should really have a separate endpoint to use here where we can apply this logic. I'm not comfortable with removing this validation, but if you are sure that it will break other parts (e.g. hrrequirement) then I will remove it.

omarabuhussein: I am of course not happy with frontend validation only , I actually do not like the way both T&A and hrrequirment depends on civicrm case tables  and many other things in other extensions and that's why I think why we should stop everything we do in civihr and sit to do some better planning because for me the whole thing is a mess and without writing some plans sooner that later nothing is going to improve.

but since I don't have the decision here I would like at least not breaking existing sutff ( which will leads to more hacks in this case) , and since there is a high rate for this modification to break thing so  I would like to avoid it .

uk.co.compucorp.civicrm.tasksassignments/CRM/Activity/Service/ActivityService.php
```diff
@@ -0,0 +1,26 @@
...
+   * Returns the activity types for a certain component. Default limit is 25
+   *
+   * @param $componentName
+   *  The name of the component
+   * @param int $limit
+   *  The maximium number of results to return
+   *
+   * @return array
+   */
+  public static function findByComponent($componentName, $limit = 0) {
```
MiyaNoctem: I think this would make the default 0, right?  So either update doc-block or the default value for $limit.

mickadoo: The default is 0 because it's used in an API call, so 0 is unlimited. I've updated the description to make this clearer

MiyaNoctem: Sorry, what I meant is that while the method's description says the default is 25, the default is unlimited, because it is set as 0.  So the method's description is not accurate, maybe better to delete "Default limit is 25", or update it to "Default limit is 0 (unlimited)".

Again, sorry for the mix-up!

mickadoo: Ah, my bad. I didn't look up enough! Changing now

PR #188, opened 2017-04-17
PCHR-2139: Cannot save joining assignment on yoti

bmnepali:
### Overview:
With heavy usage of $resource in T&A for api call, while making post requests, data is sent through query strings.

### Before:
Currently, When creating a "New Assignment with multiple tasks", Angular's `$resource` is used to make API call with post request. But the current implementation of `$resource` to post data, adds the posted data as query string in url. And in case of many tasks, the url gets much longer and such url might not get handled by server resulting `414 error`.

### Solution:
1. Add new service to handle api call with `form-urlencoded` data.
2. Modifying $resource, include headers set as `'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8'`.
3. Use service form (1) to make a api call to dave multiple tasks, sending data in request body rather than in url as query string.

### After:
1. After adding the above solution, for every post requests  for api call, data is sent through request body and `form-urlEncoded` data.

2. Also, in course of refactoring the code, "sequential: 1" is added during the initialize the $resource. i.e 
```
return $resource(config.url.REST, {
      ...
      sequential: 1,
      ...
    }
```
With this approach, as this $resource is shared across both get and post requests, the data is fetched differently when calling getOptions(). I.e with `sequential: 1` ad query string and getOptions() it returns array of objects of Task Types Option or Task Status Types, while with out it object of the respective data is returned, which was causing following issue.
screen shot 2017-04-25 at 4 27 17 pm

So, this was also fixed and added test cases for the changes.

**Important Note:** 
Also, as a part of this ticket, the path to `org.civicrm.shoreditch` theme is fixed in file `uk.co.compucorp.civicrm.tasksassignments/scss/civitasks/modules/_civihr-ui-select.scss` In Task and Assignment Module.

And ran backstopjs tests to verify everything is working fine.


bmnepali:
@AkA84 I have changed the solution, as discussed and also added some tests for the changes.

uk.co.compucorp.civicrm.tasksassignments/js/test/mocks/data/task.js
```diff
@@ -0,0 +1,72 @@
...
+        "source_contact_id": "205",
+        "status_id": "1",
+        "offset": "-3",
+        "activity_date_time": "2017-04-18",
+        "target_contact_id": [
+          "84"
+        ]
+      }
+    ],
+    remonderNote: "<p>Sample reminder note</p>"
```
AkA84: is this a typo? `remonder` instead of `reminder`?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
```diff
@@ -1,271 +1,294 @@
...
+      'action': 'get',
+      'entity': 'Task'
+    }, {
+      save: {
+        method: "POST",
+        isArray: false,
+        headers: {
+          'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8'
+        },
+        transformRequest: transformUrlEncoded
```
AkA84: just use an inline function
```js
transformRequest: function (data) {
  return $httpParamSerializer(data);
}
```
actually i wonder if this would work
```js
transformRequest: $httpParamSerializer(data);
```
can you try?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
```diff
@@ -1,271 +1,294 @@
...
+          }
+
+          var deferred = $q.defer();
+
+          Task.save({}, {
+            action: 'copy_to_assignment',
+            entity: 'Task',
+            sequential: 1,
+            debug: config.DEBUG,
+            json: JSON.stringify({
```
AkA84: what's the need for `JSON.stringify `? We weren't doing it before

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
```diff
@@ -1,271 +1,294 @@
...
+          }
+
+          if (!taskArr.length) {
+            return taskArr;
+          }
+
+          var deferred = $q.defer();
+
+          Task.save({}, {
+            action: 'copy_to_assignment',
```
AkA84: Now you're passing *all* the parameters in the request body, but i don't think that's necessary? Shouldn't you just pass `json` in the body?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
```diff
@@ -1,271 +1,294 @@
...
+              sequential: 1,
+              debug: config.DEBUG
+            }, task),
+            val;
+
+          Task.save({}, {
+            action: 'create',
+            entity: 'Task',
+            sequential: 1,
+            debug: config.DEBUG,
```
AkA84: Do `sequential: 1` and `debug: config.DEBUG` need to be within `json` (as they were before this PR)? If not, can't you just add them as default parameters when you are defining the `$resource`? They are always the same for each call

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+  describe('TaskService', function() {
+    var Task, TaskService, $provide, $httpBackend;
+
+    beforeEach(module('civitasks.appDashboard'));
+
+    beforeEach(inject(function(_Task_, _TaskService_, _$httpBackend_) {
+      Task = _Task_;
+      TaskService = _TaskService_;
+      $httpBackend = _$httpBackend_;
+      spyOn(Task, 'save').and.callThrough();
```
AkA84: Please visually group the code, it makes it more readable
```js
beforeEach(inject(function(_Task_, _TaskService_, _$httpBackend_) {
  Task = _Task_;
  TaskService = _TaskService_;
  $httpBackend = _$httpBackend_;

  spyOn(Task, 'save').and.callThrough();
}));
```

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+            source_contact_id: '205',
+            target_contact_id: ['10'],
+            details: '<p>Detailed description of the task</p>',
+            case_id: null,
+            activity_type_id: '88',
+            subject: 'Sample Task Subject'
+          })
+        };
+
+        expect(Task.save.calls.argsFor(0)[1]).toEqual(params);
```
AkA84: I think you are mixing testing two different things here, `Task` (the `$resource`) and `TaskService` (the service that uses the `$resource`). Unfortunately they're kept in the same file, so for now you can simply have two main `describe` blocks, where you test the two things separately

```js
describe('Task', function () {
  // test the resource
});

describe('TaskService', function () {
  // Test the service
});
```

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+            sequential: 1,
+            debug: true,
+            activity_date_time: '2017-04-21',
+            assignee_contact_id: ['6'],
+            source_contact_id: '205',
+            target_contact_id: ['10'],
+            details: '<p>Detailed description of the task</p>',
+            case_id: null,
+            activity_type_id: '88',
+            subject: 'Sample Task Subject'
```
AkA84: You are using mocked data
```js
$httpBackend.whenPOST(/rest&action=get&entity=Task/).respond({
  is_error: 0,
  values: [TaskMock.task]
});
```
so there is no need for you to hardcode these values. It means that if for some reason you are going to update the mocked data, this test will fail because the values won't match anymore.

Instead just reference that mocked data directly

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
AkA84: Make the tests more expressive, testing for explicit bits of data you are really interested in, instead of just generally testing for the whole object
```js
var requestBody = Task.save.calls.argsFor(0)[1];

expect(requestBody).toEqual(jasmine.objectContaining({
  action: 'create',
  entity: 'Task',
  json: // the mocked data
}));
```

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+        });
+        TaskService.save(TaskMock.task);
+        $httpBackend.flush();
+      });
+
+      it('makes call to save()', function() {
+        expect(Task.save).toHaveBeenCalled();
+      });
+
+      it('calls save() with correct data', function() {
```
AkA84: this is a test for `Task`

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+            case_id: null,
+            activity_type_id: '88',
+            subject: 'Sample Task Subject'
+          })
+        };
+
+        expect(Task.save.calls.argsFor(0)[1]).toEqual(params);
+      });
+
+      it("calls request with correct Content-Type headers", function() {
```
AkA84: this is a test for `TaskService`

AkA84: Also, don't just say the "correct headers", explicitly say what these headers are

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+        $httpBackend.whenPOST(/rest&action=get&entity=Task/)
```
AkA84: I think there are some better ways to test for request headers, please check here for example http://stackoverflow.com/questions/43478186/how-to-test-resource-calls-if-it-contains-specific-headers-on-making-http-reque

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+          .respond(function(method, url, data, headers, params) {
+            expect(headers['Content-Type']).toEqual('application/x-www-form-urlencoded; charset=UTF-8');
+            return {};
+          });
+
+        TaskService.save(TaskMock.task);
+        $httpBackend.flush();
+      });
+
+      it("adds url encodning to task lists", function() {
```
AkA84: this is a test for `TaskService`

AkA84: also, typo

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+        $httpBackend.whenPOST(/rest&action=get&entity=Task/)
```
AkA84: same as before, isn't there a better way to test this?

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+      beforeEach(function() {
+        $httpBackend.whenPOST(/rest&action=get&entity=Task/).respond({
+          is_error: 0,
+          values: [TaskMock.task]
+        });
+        TaskService.assign(TaskMock.taskList, 1);
+        $httpBackend.flush();
+      });
+
+      it('calls to $resource\'s save()', function() {
```
AkA84: just use `"` in this case, to avoid the escaping character

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+          json: JSON.stringify({
+            "id": TaskMock.taskList,
+            "case_id": 1
+          })
+        };
+
+        expect(Task.save.calls.argsFor(0)[1]).toEqual(params);
+      });
+
+      it("calls request with corect Content-Type header", function() {
```
AkA84: typo

AkA84: also, you already test this, no point in testing it again

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+          .respond(function(method, url, data, headers, params) {
+            expect(headers['Content-Type']).toEqual('application/x-www-form-urlencoded; charset=UTF-8');
+            return {};
+          });
+
+        TaskService.assign(TaskMock.taskList, 1); // 1 =>addignmentId
+        $httpBackend.flush();
+      });
+
+      it("adds url encodning to task list", function() {
```
AkA84: you already test this, no point in testing it again

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+          debug: true,
+          json: JSON.stringify({
+            "task": TaskMock.taskList
+          })
+        };
+
+        expect(Task.save.calls.argsFor(0)[1]).toEqual(params);
+      });
+
+      it("calls request with Content-Typeheader", function() {
```
AkA84: also, you already test this, no point in testing it again

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+          .respond(function(method, url, data, headers, params) {
+            expect(headers['Content-Type']).toEqual('application/x-www-form-urlencoded; charset=UTF-8');
+            return {};
+          });
+
+        TaskService.saveMultiple(TaskMock.taskList);
+        $httpBackend.flush();
+      });
+
+      it("adds url encodning to task lists", function() {
```
AkA84: also, you already test this, no point in testing it again

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+            "notes": TaskMock.remonderNode
+          })
+        };
+
+        expect(Task.save.calls.argsFor(0)[1]).toEqual(params);
+        TaskService.sendReminder(1, TaskMock.remonderNode);
+        $httpBackend.flush();
+      });
+
+      it("calls request with Content-Type header", function() {
```
AkA84: also, you already test this, no point in testing it again

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,246 @@
...
+          .respond(function(method, url, data, headers, params) {
+            expect(headers['Content-Type']).toEqual('application/x-www-form-urlencoded; charset=UTF-8');
+            return {};
+          });
+
+        TaskService.sendReminder(1, TaskMock.remonderNode);
+        $httpBackend.flush();
+      });
+
+      it("adds url encodning to notes data", function() {
```
AkA84: also, you already test this, no point in testing it again

uk.co.compucorp.civicrm.tasksassignments/js/test/mocks/data/task.js
bmnepali: ops! will fix it.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
bmnepali: sure!

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
bmnepali: No it doesn't need to, rather its appropriate to send in request body. I prefer to add only required data as query string that is `action` and `entity`, and send every other data in request body.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
AkA84: really sorry, i meant
```js
transformRequest: $httpParamSerializer
```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
AkA84: if they don't need to, then it means that you can put them in the params object (that is, the first argument), and set them as default when you define the $resource
```js
return $resource(config.url.REST, {
  action: 'get',
  entity: 'Task',
  sequential: 1,
  debug: config.DEBUG
}, {
  save: { /* .. */ }
});
```

this way you can remove duplication, as you can get rid of those two in every call to `Task`

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
```diff
@@ -216,14 +207,12 @@ define([
 
           var deferred = $q.defer();
 
-          Task.save({}, {
-            action: 'create_multiple',
-            entity: 'Task',
+          Task.save({action: 'create_multiple'}, {
```
AkA84:
```js
Task.save({ action: 'create_multiple' }, {
```

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+        beforeEach(function() {
+          $httpBackend.whenPOST(/action=create&entity=Task/).respond(function(method, url, data, headers, params) {
+            request.method  = method;
+            request.headers = headers;
+            request.data = data;
+
+            return [200];
+          });
+
+          TaskService.save(TaskMock.task);
```
AkA84: you are testing *only* `Task` here, `TaskService` should not be used at all. In unit tests you test only one thing at a time and you mock any dependency. `TaskService` here it's not even a dependency, it's just a service that uses `Task` (so `Task` is a dependency for `TaskService`)

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+    beforeEach(function() {
+      $httpBackend.whenGET(/action=getoptions&entity=Task/).respond({});
+      $httpBackend.whenGET(/action=getoptions&entity=Document/).respond({});
+      $httpBackend.whenGET(/action=get&entity=CaseType/).respond({});
+      $httpBackend.whenGET(/action=get&entity=contact/).respond({});
+      $httpBackend.whenGET(/action=get&entity=Task/).respond({});
+      $httpBackend.whenGET(/views.*/).respond({});
+    });
+
+    describe('$resource.save()', function() {
```
AkA84:
```js
describe('save()', function () {
});
```

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+
+            return [200];
+          });
+
+          TaskService.save(TaskMock.task);
+          $httpBackend.flush();
+        });
+
+        it('makes call to save()', function() {
+          expect(Task.save).toHaveBeenCalled();
```
AkA84: as said before, you cannot use `TaskService` while testing `Task`, you should test `Task` directly

uk.co.compucorp.civicrm.tasksassignments/js/test/mocks/data/task.js
```diff
@@ -0,0 +1,184 @@
+define(function() {
+  return {
+    task: {
```
AkA84: I think there is a big misunderstanding as to what this file is supposed to hold. The mocked data should be the representation of the data *coming from the backend API*, that's all. You should simply have a static, mocked data that you would get in reality from the API.

In here instead i see that you have an example task object, some url encoded data, an array with some tasks objects, etc. They obviously have nothing to do with what you would get *from* the API.

If you want to have a quick way to generate tasks on the fly for testing purposes, then you could think of having a `test/fabricators` folder where you could have `task.js`, `assignment.js`.

Just for not getting the scope of this PR too big, for now `test/fabricators/task.js` could simply return those static object & array you have currently in this file
```js
// test/fabricators/task.js

define([], function () {
  return {

    /**
     * STATIC FOR NOW!
     * 
     * [task description]
     * @return {[type]} [description]
     */
    single: function () {
        return {
          activity_date_time: "2017-04-21",
          assignee_contact_id: ["6"],
          source_contact_id: "205",
          target_contact_id: ["10"],
          details: "<p>Detailed description of the task</p>",
          case_id: null,
          activity_type_id: "88",
          subject: "Sample Task Subject"
        };
    },

    /**
     * STATIC FOR NOW!
     *
     * [tasksList description]
     * @return {[type]} [description]
     */
    list: function () {
      return /* the list you currently have in mock/data/task */
    }
  }
};
```

then in the actual test file you could simply require it and then use it for whatever you need
```js
define([
  test/mocks/fabricators/task
], function (taskFabricator) {
  var fakeTask = taskFabricator.single();
  var fakeTaskList  = taskFabricator.list();
});
```

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+      $httpBackend.whenGET(/action=get&entity=CaseType/).respond({});
+      $httpBackend.whenGET(/action=get&entity=contact/).respond({});
+      $httpBackend.whenGET(/action=get&entity=Task/).respond({});
+      $httpBackend.whenGET(/views.*/).respond({});
+    });
+
+    describe('$resource.save()', function() {
+      describe('when calling TaskService.save()', function () {
+        beforeEach(function() {
+          $httpBackend.whenPOST(/action=create&entity=Task/).respond(function(method, url, data, headers, params) {
```
AkA84: you don't need `params`

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+        it('makes call to save()', function() {
+          expect(Task.save).toHaveBeenCalled();
+        });
+
+        it('makes request with Content-Type as application/x-www-form-urlencoded; charset=UTF-8', function() {
+          expect(request.headers['Content-Type']).toEqual('application/x-www-form-urlencoded; charset=UTF-8');
+        });
+
+        it('adds url encoding to Task list', function () {
+          expect(request.data).toEqual(TaskMock.urlEncodedData.onSave);
```
AkA84: the url encoded data should not be in the mocked data (see my previous comment). You can simply manually create it from the task mock, and then compare it with `request.data`

```js
beforeEach(function () {
  var fakeTask = taskFabricator.single();
  var expectedEncodedData = /* encode the data of fakeTask */
});

it('url encodes the task data', function () {
  expect(request.data).toBe(expectedEncodedData);
});
```

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+      describe("when calling TaskService.assign()", function() {
+        beforeEach(function() {
+          $httpBackend.whenPOST(/action=copy_to_assignment&entity=Task/).respond(function(method, url, data, headers, params) {
+            request.headers = headers;
+            request.data = data;
+
+            return [200];
+          });
+
+          TaskService.assign(TaskMock.taskList, 1);
```
AkA84: same thing as before, you should test `Task` directly

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+              case_id: 1
+            }
+          };
+          requestBody = Task.save.calls.argsFor(0)[1];
+
+          expect(requestBody.json).toEqual(params.json);
+        });
+      });
+
+      describe("when calling TaskService.saveMultiple()", function() {
```
AkA84: This whole main `describe` block is for `Task`, not  `TaskService`

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+              task: TaskMock.taskList
+            }
+          };
+          requestBody = Task.save.calls.argsFor(0)[1],
+
+          expect(requestBody.json).toEqual(params.json);
+        });
+      });
+
+      describe("when calling TaskService.sendReminder()", function() {
```
AkA84: see above

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+
+    beforeEach(module('civitasks.appDashboard'));
+
+    beforeEach(inject(function(_TaskService_, _$httpBackend_) {
+      TaskService = _TaskService_;
+      $httpBackend = _$httpBackend_;
+    }));
+
+    beforeEach(function() {
+      $httpBackend.whenGET(/action=getoptions&entity=Task/).respond({});
```
AkA84: duplicated code, put it in a function

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+    });
+
+    describe("save()", function() {
+      beforeEach(function() {
+        $httpBackend.whenPOST(/action=create&entity=Task/).respond(function() {
+          return [200, TaskMock.response.onSave];
+        });
+      });
+
+      it('responds correct Task details', function (){
```
AkA84: i don't undertand what you're testing here, the description is not clear, could you explain it to me?

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+    });
+
+    describe("assign()", function() {
+      beforeEach(function() {
+        $httpBackend.whenPOST(/action=copy_to_assignment&entity=Task/).respond(function() {
+          return [200, TaskMock.response.onAssign];
+        });
+      });
+
+      it('responds assigned Task List', function (){
```
AkA84: i don't undertand what you're testing here, the description is not clear, could you explain it to me?

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+    describe("saveMultiple()", function() {
+      beforeEach(function() {
+        $httpBackend.whenPOST(/action=create_multiple&entity=Task/).respond(function() {
+          return [200, TaskMock.response.onSaveMultiple];
+        });
+      });
+
+      it('responds saved Task List', function (){
+        TaskService.saveMultiple(TaskMock.taskList).then(function(data){
+          expect(data).toEqual(TaskMock.response.onSaveMultiple.values);
```
AkA84: you are manually setting up the backend to respond with `TaskMock.response.onSaveMultiple`
```js
beforeEach(function() {
  $httpBackend.whenPOST(/action=create_multiple&entity=Task/).respond(function() {
    return [200, TaskMock.response.onSaveMultiple];
  });
});
```
so how can 
```js
expect(data).toEqual(TaskMock.response.onSaveMultiple.values);
```
ever fails? You are manually setting the test to always respond with that data, this test cannot ever fail, hence it's not a meaningful test

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,225 @@
...
+    describe("sendReminder()", function() {
+      beforeEach(function() {
+        $httpBackend.whenPOST(/action=sendreminder&entity=Task/).respond(function() {
+          return [200, TaskMock.response.onSendReminder];
+        });
+      });
+
+      it('responds with reminder sent status', function (){
+        TaskService.sendReminder(1, TaskMock.reminderNote).then(function(data){
+          expect(data.values).toEqual(TaskMock.response.onSendReminder.values);
```
AkA84: same argument as before

uk.co.compucorp.civicrm.tasksassignments/js/test/mocks/data/task.js
```diff
@@ -0,0 +1,135 @@
+define(function () {
+  return {
+    response: function () {
```
AkA84: why do you need `response` at all? Can't you simply have
```js
return {
  onAssign: function () { },
  onSave: function () { },
  // etc
}

File: uk.co.compucorp.civicrm.tasksassignments/js/test/mocks/fabricators/task.js [hunk 37]
```diff
@@ -0,0 +1,74 @@
+define(function () {
+  return {
+    single: function () {
+        return {
```
AkA84: wrong indentation

uk.co.compucorp.civicrm.tasksassignments/js/test/mocks/fabricators/task.js
```diff
@@ -0,0 +1,74 @@
...
+          assignee_contact_id: ["6"],
+          source_contact_id: "205",
+          target_contact_id: ["10"],
+          details: "<p>Detailed description of the task</p>",
+          case_id: null,
+          activity_type_id: "88",
+          subject: "Sample Task Subject"
+        };
+    },
+
```
AkA84: empty line

uk.co.compucorp.civicrm.tasksassignments/js/test/mocks/fabricators/task.js
```diff
@@ -0,0 +1,74 @@
...
+          "status_id": "1",
+          "offset": "-3",
+          "activity_date_time": "2017-04-18",
+          "target_contact_id": [
+            "84"
+          ]
+        }
+      ]
+    },
+
```
AkA84: empty line

uk.co.compucorp.civicrm.tasksassignments/js/test/mocks/data/task.js
bmnepali: right! will do.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
```diff
@@ -116,10 +116,10 @@ define([
               taskType.arr.push({
                 key: data.values[optionId].key,
                 value: data.values[optionId].value
-              })
-            }
+              });
```
deb1990: You can avoid unnecessary variable creation and nested values fetching by this
```
_.each(data.values, function(option) {
  taskType.arr.push({
    key: option.key,
    value: option.value
  });

  taskType.obj[option.key] = option.value;
})
```
Same should be done below.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
```diff
@@ -134,12 +134,12 @@ define([
...
-                key: optionId,
-                value: data.values[optionId]
-              })
-            }
+                key: data.values[optionId].key,
+                value: data.values[optionId].value
+              });
 
-            taskStatus.obj = data.values;
+              taskStatus.obj[data.values[optionId].key] = data.values[optionId].value;
```
deb1990: Test case should be updated

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -73,13 +75,16 @@ define([
   });
 
   describe('TaskService', function () {
-    var TaskService, $httpBackend;
+    var Task,TaskService, $httpBackend, $q, deferred;
```
deb1990: Space after Task,

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -73,13 +75,16 @@ define([
...
 
     beforeEach(module('civitasks.appDashboard'));
 
-    beforeEach(inject(function (_TaskService_, _$httpBackend_) {
+    beforeEach(inject(function ( _TaskService_, _$httpBackend_, _$q_) {
```
deb1990: No space before _TaskService_

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -73,13 +75,16 @@ define([
...
       TaskService = _TaskService_;
       $httpBackend = _$httpBackend_;
+      $q = _$q_;
+      deferred = _$q_.defer();
+
```
deb1990: Empty line

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -73,13 +75,16 @@ define([
...
+    var Task,TaskService, $httpBackend, $q, deferred;
 
     beforeEach(module('civitasks.appDashboard'));
 
-    beforeEach(inject(function (_TaskService_, _$httpBackend_) {
+    beforeEach(inject(function ( _TaskService_, _$httpBackend_, _$q_) {
       TaskService = _TaskService_;
       $httpBackend = _$httpBackend_;
+      $q = _$q_;
+      deferred = _$q_.defer();
```
deb1990: `$q.defer`

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -149,5 +158,45 @@ define([
...
+
+    describe("getActivityTypeIdOption()", function () {
+      beforeEach(function () {
+        $httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond(function () {
+          return [200, taskMock.onGetOptions.activityTypeIdFields];
+        });
+      });
+
+      it("returns promise with array of list of activity types options", function () {
+        $q.all({
```
deb1990: It should contain only the `expect`and other unavoidable statements. You should be aware of that by now. 

```javascript


var promise;

beforeEach(function() {
	promise = $q.all({
	  taskType: TaskService.getActivityTypeIdOption()
	});
});

afterEach(function() {
	$httpBackend.flush();
});

it("returns promise with array of list of activity types options", function () {
	promise.then(function(options) {
	  expect(options.taskType.obj).toBeDefined();
	  expect(options.taskType.arr).toEqual(fakeActivityTypesOptions);
	});
});
```

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -149,5 +158,45 @@ define([
...
+          taskType: TaskService.getActivityTypeIdOption()
+        }).then(function(options) {
+          expect(options.taskType).toBeDefined();
```
deb1990: Unnecessary check as you are already checking `options.taskType.obj`

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -149,5 +158,45 @@ define([
...
+    describe("getStatusIdOption()", function () {
+      beforeEach(function () {
+        $httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond(function () {
+          return [200, taskMock.onGetOptions.statusIdFields];
+        });
+      });
+
+      it("returns promise with array of list of statys types options", function () {
+        $q.all({
+          taskStatus: TaskService.getStatusIdOption()
```
deb1990: Same as above.

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -149,5 +158,45 @@ define([
         $httpBackend.flush();
       });
     });
+
+    describe("getActivityTypeIdOption()", function () {
+      beforeEach(function () {
+        $httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond(function () {
```
deb1990:
```
$httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond(taskMock.onGetOptions.activityTypeIdFields);
```

Also check other similar instances

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,222 @@
...
+        $httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond({});
+        $httpBackend.whenPOST(/action=sendreminder&debug=true&entity=Task/).respond(taskMock.onSendReminder);
+      });
+
+      it('returns reminder sent status on sending reminder for a task', function () {
+        TaskService.sendReminder(1, fakeReminderNote).then(function (data) {
+          expect(data.values).toEqual(true);
+        });
+
+        $httpBackend.flush();
```
deb1990: It should be in after Each

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,222 @@
...
+        $httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond({});
+        $httpBackend.whenPOST(/action=create_multiple&debug=true&entity=Task/).respond(taskMock.onSaveMultiple);
+      });
+
+      it('returns saved task list on saving multiple tasks', function () {
+        TaskService.saveMultiple(fakeTaskList).then(function (data) {
+          expect(data).toEqual(fakeTaskList);
+        });
+
+        $httpBackend.flush();
```
deb1990: It should be in after Each

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,222 @@
...
+        $httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond({});
+        $httpBackend.whenPOST(/action=copy_to_assignment&debug=true&entity=Task/).respond(taskMock.onAssign);
+      });
+
+      it('returns list of assigned tasks on task assign', function () {
+        TaskService.assign(fakeTaskList, 1).then(function (data) {
+          expect(data).toEqual(fakeTaskList);
+        });
+
+        $httpBackend.flush();
```
deb1990: It should be in after Each

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,222 @@
...
+        $httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond({});
+        $httpBackend.whenPOST(/action=create&debug=true&entity=Task/).respond(taskMock.onSave);
+      });
+
+      it('returns saved Task on saving task', function () {
+        TaskService.save(fakeTask).then(function (data) {
+          expect(data).toEqual(fakeTask);
+        });
+
+        $httpBackend.flush();
```
deb1990: It should be in after Each

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,222 @@
...
+
+    beforeEach(function () {
+      $httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond({});
+      mockBackendCalls($httpBackend);
+    });
+
+    describe('save()', function () {
+      beforeEach(function () {
+        $httpBackend.whenPOST(/action=create&debug=true&entity=Task/).respond(function (method, url, data, headers) {
+          request.method = method;
```
deb1990: Why do you need to set all this,
request.method = method;
request.headers = headers;
request.data = data;
return [200];

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,222 @@
...
+      });
+
+      it('adds url encoding to Task list', function () {
+        expect(request.data).toEqual($httpParamSerializer({
+          json: fakeTask
+        }));
+      });
+
+      it('calls save() as POST request with Task data', function () {
+        requestBody = Task.save.calls.argsFor(0)[1];
```
deb1990: Create a describe and place this into beforeEach. 
**Any kind of setup for a test case other than than expect should go to beforeEach.
Any kind of code which should run after expect should go to afterEach.**

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,222 @@
...
+
+        expect(request.method).toEqual('POST');
+        expect(requestBody.json).toEqual(fakeTask);
+      });
+    });
+  });
+
+  describe('TaskService', function () {
```
deb1990: This is another File you are testing, new file should be created for this

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
```diff
@@ -0,0 +1,222 @@
...
+        $httpBackend.whenGET(/action=getoptions&debug=true&entity=Task/).respond(taskMock.onGetOptions.taskStatus);
+      });
+
+      beforeEach(function () {
+        promise = $q.all({
+          taskStatus: TaskService.getTaskStatus()
+        });
+      });
+
+      afterEach(function () {
```
deb1990: Please check if `$httpBackend.flush();` is called for every test case, if yes then place it in the outer describe's afterEach and remove individual one's

bmnepali: right! move it to outer describe block.

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
bmnepali: I feel its more appropriate to keep the Task factory within this file task.js itself, as in the Task factory, we are only initialiazing the $resource + overriding $resource.save() function specific to task get and post calls.

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
bmnepali: ^ these are simply to test if the request is made with appropriate method, headers and data

uk.co.compucorp.civicrm.tasksassignments/js/test/services/task_test.js
bmnepali: Got it! will refactor the code to address this issue.

bmnepali:
Fixed merge conflicts!

bmnepali:
Fixed merge conflicts!

PR #189, opened 2017-04-20
PCHR-2151: New Fields for Documents

mickadoo:
## Overview
Adds three new fields for documents

- Valid from
- Document Number
- Remind Me

## Before
The only custom fields for documents were expire_date and merge_date.

## After
The three new fields will be added and available in the API explorer.

## Technical Details
The new document fields were added manually and exported using `civix generate:custom-xml --data=14`. I added the new fields to the existing XML since existing fields will be ignored.

It was decided that all existing documents should have "remind me" set to true, so I've added an upgrader for this.

## Comments
Originally "document number" was due to be called "document ID" but since this clashes with the actual ID field I changed it.

This PR targets PCHR-2145 branch as it is part of a larger ticket which will be tested as a whole when all subtickets are complete.

---

- [ ] Tests Pass

I do not have testing set up locally

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -577,8 +577,7 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
    *
    * @return bool
    */
-  public function upgrade_1025()
-  {
+  public function upgrade_1025() {
     $this->executeCustomDataFile('xml/activity_custom_fields.xml');
 
     return TRUE;
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -580,6 +580,12 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
   public function upgrade_1025() {
     $this->executeCustomDataFile('xml/activity_custom_fields.xml');
 
+    // set "remind_me" to true for all existing documents
+    civicrm_api3('Document', 'get', [
+      'options' => ['limit' => 0],
+      'api.Document.create' => ['id' => '$value.id', 'remind_me' => 1],
+    ]);
+
     return TRUE;
   }
 
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -460,7 +460,7 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
    * Install Tasks Assignments 'days_to_create_a_document_clone' setting field.
    * It keeps a number of days to create a document clone before original
    * expiry date.
-   * 
+   *
    * @return {boolean}
    */
   public function upgrade_1020() {
@@ -513,7 +513,7 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
   /*
    * Set up scheduled job which clones documents on pre-set days before
    * their original expiry date.
-   * 
+   *
    * @see PCHR-1365
    */
   public function upgrade_1022()
@@ -572,6 +572,18 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
     return TRUE;
   }
 
+  /**
+   * Add new custom fields for activities
+   *
+   * @return bool
+   */
+  public function upgrade_1025()
+  {
+    $this->executeCustomDataFile('xml/activity_custom_fields.xml');
+
+    return TRUE;
+  }
+
     public function uninstall()
     {
         CRM_Core_DAO::executeQuery("DELETE FROM `civicrm_navigation` WHERE name IN ('tasksassignments', 'ta_dashboard', 'tasksassignments_administer', 'ta_settings')");
@@ -585,7 +597,7 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
      * specifically it returns the option group and component id
      *
      * @param  string $component
-     * @return Array
+     * @return array
      */
     private function _fetchActivityTypeParams($component) {
       $componentId = null;
--- a/uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
@@ -27,29 +27,11 @@ function civicrm_api3_document_create($params) {
     );
   }
 
-  /* This code prevents creating a copy of Activity each time we pass 'case_id'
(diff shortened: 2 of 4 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -572,6 +572,18 @@ public function upgrade_1024()
     return TRUE;
   }
 
+  /**
+   * Add new custom fields for activities
+   *
+   * @return bool
+   */
+  public function upgrade_1025()
+  {
```
tunbola: Opening curly brace should be on same line as function name according to Drupal coding standards

uk.co.compucorp.civicrm.tasksassignments/xml/activity_custom_fields.xml
```diff
@@ -53,7 +54,70 @@
...
+      <note_columns>60</note_columns>
+      <note_rows>4</note_rows>
+      <column_name>document_id</column_name>
+      <in_selector>0</in_selector>
       <custom_group_name>Activity_Custom_Fields</custom_group_name>
     </CustomField>
   </CustomFields>
-</CustomData>
\ No newline at end of file
+</CustomData>
```
tunbola: There should be a new line at the end of this file since we are following Drupal's coding convention.
BTW, your IDE should add that automatically?

uk.co.compucorp.civicrm.tasksassignments/xml/activity_custom_fields.xml
```diff
@@ -53,7 +54,70 @@
...
+    </CustomField>
+    <CustomField>
+      <name>Valid_From</name>
+      <label>Valid From</label>
+      <data_type>Date</data_type>
+      <html_type>Select Date</html_type>
+      <is_required>0</is_required>
+      <is_searchable>1</is_searchable>
+      <is_search_range>1</is_search_range>
+      <weight>19</weight>
```
tunbola: Any reason why the weight for the new custom fields you added does not sequentially follow the ones already added in this file?

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
```diff
@@ -232,32 +214,43 @@ function civicrm_api3_document_create_multiple($params) {
  */
 function _civicrm_api3_document_create_spec(&$params) {
 
-  //default for source_contact_id = currently logged in user
-  $params['source_contact_id']['api.default'] = 'user_contact_id';
```
tunbola: Do we really need to remove this? I know civi defaults to current user on the API explorer for most of its API's that uses a contact ID.

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
```diff
@@ -232,32 +214,43 @@ function civicrm_api3_document_create_multiple($params) {
...
       'title' => 'Activity Source Contact',
       'type' => 1,
       'FKClassName' => 'CRM_Activity_DAO_ActivityContact',
       'api.default' => 'user_contact_id',
-  );
+  ];
 
+  // replace custom data options with their column name
+  foreach ($params as $key => $param) {
+    if (substr($key, 0, 7) === 'custom_' && isset($param['column_name'])) {
```
tunbola: is this logic fail safe? I mean will civi always name the keys to begin with 'custom_'. I inspected the array on my local and see that there is an extend property with has the value:  'Activity' for the custom fields.

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
```diff
@@ -609,22 +600,21 @@ function civicrm_api3_document_get($params) {
...
+            $customFields = _civicrm_api3_document_getcustomfields();
             foreach ($customFields as $customFieldKey => $customFieldData) {
-                $customColumnResult = CRM_Core_DAO::executeQuery(
-                    "SELECT {$customFieldData['name']} FROM {$customFieldData['table']} WHERE entity_id = %1",
-                    array(
-                        1 => array($getResult['values'][$key]['id'], 'Integer'),
-                    )
-                );
-                if (!$customColumnResult->fetch()) {
+
```
tunbola: Nice optimization here.

tunbola:
Also in the PR description, 'The only custom fields for documents were expire_date and merge_date.'
I checked and it seems the previous custom fields are expire_date and clone_date.

uk.co.compucorp.civicrm.tasksassignments/xml/activity_custom_fields.xml
mickadoo: No, just I used the autogenerated weights from the exported xml. I'll change them to sequential.

uk.co.compucorp.civicrm.tasksassignments/xml/activity_custom_fields.xml
mickadoo: There should be a newline there, on the previous file I see an icon in the comparison saying "no new line at end of file"

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
mickadoo: I removed it because it's duplicated in a block a few lines below `$params['source_contact_id'] = [`

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
mickadoo: discussed this point with @tunbola in chat, an extra check if `$param['extends'] === 'Activity'` would add an extra layer of safety. It may not be required, but potentially could be a case where activity has a real field starting with "custom_" where we don't want this logic applied.

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
```diff
@@ -241,18 +242,30 @@ function _civicrm_api3_document_create_spec(&$params) {
...
+    'options' => ['limit' => 0],
+  ])['values'];
 
-    $parent = CRM_Utils_Array::value('extends', $param);
+  foreach ($params as $key => $param) {
+    $parent = ArrayHelper::value('extends', $param);
     $isCustom = substr($key, 0, 7) === 'custom_';
 
     if ($isCustom && $parent === 'Activity' && isset($param['column_name'])) {
+      $customFieldId = substr($key, strrpos($key, '_') + 1);
```
tunbola: Couldn't understand what was happening here at a glance. Maybe a comment like:
//For example custom_85 has an ID of 85 as a custom field.
What do you think?

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
```diff
@@ -241,18 +242,30 @@ function _civicrm_api3_document_create_spec(&$params) {
       'api.default' => 'user_contact_id',
   ];
 
-  // replace custom data options with their column name
-  foreach ($params as $key => $param) {
+  $metadata = civicrm_api3('CustomField', 'get', [
+    'custom_group_id' => "Activity_Custom_Fields",
```
tunbola: Please use single quotes here

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
mickadoo: Sounds good, I'll add a comment `// custom field names have the format custom_<id>`

PR #192, opened 2017-04-24
PCHR-2152: Remind Me Functionality

mickadoo:
## Overview
The functionality for cloning documents that are approaching their expiry date is already developed and in use. After adding the "remind me" column in #189 we need to check before cloning if the user has selected that they want to be reminded about the document before cloning it.

## Before
All documents that had an expiry date soon approaching were cloned.

## After
Only documents with "remind me" set to true will be cloned (if their expiry date is approaching).

## Technical Details
- Originally I had to unset all the custom field names (there are 3 variations for each e.g. `expire_date`, `custom_77` & `custom_77_9`) before creating but by using "return" in the get request it limits the field names.

## Comments
- I tried to tidy up by removing raw SQL usage

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
@@ -77,6 +77,14 @@ class CRM_Tasksassignments_BAO_Document extends CRM_Tasksassignments_DAO_Documen
     $clonedField = static::getCustomFieldName('clone_date');
     $remindMeField = static::getCustomFieldName('remind_me');
 
+    $fieldsRequiredForClone = [
+      'target_contact_id',
+      'activity_type_id',
+      'details',
+      'expire_date',
+      'activity_date_time',
+    ];
+
     $params = [
       $expiryField => ['<=' => $cutOffDate->format('Y-m-d')],
       $clonedField => ['IS NULL' => 1],
@@ -84,7 +92,7 @@ class CRM_Tasksassignments_BAO_Document extends CRM_Tasksassignments_DAO_Documen
       'is_deleted' => 0,
       'status_id' => self::STATUS_APPROVED,
       'options' => ['limit' => 0],
-      'return' => ['target_contact_id', 'activity_type_id', 'details']
+      'return' => $fieldsRequiredForClone
     ];
 
     $result = civicrm_api3('Document', 'get', $params);
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
@@ -82,6 +82,7 @@ class CRM_Tasksassignments_BAO_Document extends CRM_Tasksassignments_DAO_Documen
       'is_deleted' => 0,
       'status_id' => self::STATUS_APPROVED,
       'options' => ['limit' => 0],
+      'return' => ["target_contact_id", "activity_type_id", "details"]
     ];
 
     $result = civicrm_api3('Document', 'get', $params);
@@ -97,18 +98,16 @@ class CRM_Tasksassignments_BAO_Document extends CRM_Tasksassignments_DAO_Documen
   protected static function cloneDocument(array $original) {
     $clone = $original;
     $origId = $original['id'];
-    $expiryField = static::getCustomFieldName('expire_date');
 
     $clone['activity_date_time'] = $original['expire_date'];
     $clone['status_id'] = self::STATUS_AWAITING_UPLOAD;
 
-    $fields = [$expiryField, 'expire_date', 'file_count', 'id'];
-    $fields[] = sprintf('%s_%s', $expiryField, $origId); // alternate name
-    foreach ($fields as $field) {
+    $fieldsToUnset = ['id', 'file_count', 'expire_date', 'document_id', 'valid_from'];
+    foreach ($fieldsToUnset as $field) {
       unset($clone[$field]);
     }
 
-    civicrm_api3('Document', 'create', $original);
+    civicrm_api3('Document', 'create', $clone);
 
     // Update original's clone_date' so we know it has been cloned.
     $today = (new DateTime())->format('Y-m-d');
@@ -131,6 +130,7 @@ class CRM_Tasksassignments_BAO_Document extends CRM_Tasksassignments_DAO_Documen
       array_walk($fieldNames, function (&$customField) {
         $customField = $customField['name'];
       });
+      $fieldNames = array_flip($fieldNames);
     }
 
     return CRM_Utils_Array::value($fieldName, $fieldNames);
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
@@ -83,6 +83,7 @@ class CRM_Tasksassignments_BAO_Document extends CRM_Tasksassignments_DAO_Documen
       'details',
       'expire_date',
       'activity_date_time',
+      'assignee_contact_id',
     ];
 
     $params = [
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
@@ -11,7 +11,7 @@ class CRM_Tasksassignments_BAO_Document extends CRM_Tasksassignments_DAO_Documen
    * Create a new Document based on array-data
    *
    * @param array $params key-value pairs
-   * @return CRM_Tasksassignments_DAO_Document|NULL
+   * @return CRM_Tasksassignments_DAO_Document|NULL|object
    */
   public static function create(&$params) {
     $entityName = 'Document';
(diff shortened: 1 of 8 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -11,7 +11,7 @@ class CRM_Tasksassignments_BAO_Document extends CRM_Tasksassignments_DAO_Documen
    * Create a new Document based on array-data
    *
    * @param array $params key-value pairs
-   * @return CRM_Tasksassignments_DAO_Document|NULL
```
davialexandre: Why did you need to add the object type? Note that we already say this method returns a Document DAO instance

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -38,118 +38,115 @@ public static function create(&$params) {
...
 
   /**
-   * Return a 'days_to_create_a_document_clone' value from T&A Settings.
-   * 
-   * @return int
+   * Finds all documents that should be cloned
+   *
+   * @param DateTime $cutOffDate
+   *  Documents with an expiry date less than this will be cloned.
+   * @return array
```
davialexandre: Please add an empty line between the `@param` and `@return` sections

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -38,118 +38,115 @@ public static function create(&$params) {
...
   }
 
   /**
-   * Return a 'days_to_create_a_document_clone' value from T&A Settings.
-   * 
-   * @return int
+   * Finds all documents that should be cloned
+   *
+   * @param DateTime $cutOffDate
+   *  Documents with an expiry date less than this will be cloned.
```
davialexandre: Here you say less than, but on the API call you use `<=`. Which one is correct?

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -38,118 +38,115 @@ public static function create(&$params) {
...
+    $remindMeField = static::getCustomFieldName('remind_me');
+
+    $params = [
+      $expiryField => ['<=' => $cutOffDate->format('Y-m-d')],
+      $clonedField => ['IS NULL' => 1],
+      $remindMeField => 1,
+      'is_deleted' => 0,
+      'status_id' => self::STATUS_APPROVED,
+      'options' => ['limit' => 0],
+      'return' => ["target_contact_id", "activity_type_id", "details"]
```
davialexandre: please use single quotes for the field names

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -38,118 +38,115 @@ public static function create(&$params) {
...
+    $fieldsToUnset = ['id', 'file_count', 'expire_date', 'document_number', 'valid_from'];
+    foreach ($fieldsToUnset as $field) {
+      unset($clone[$field]);
     }
-    return $count;
+
+    civicrm_api3('Document', 'create', $clone);
+
+    // Update original's clone_date' so we know it has been cloned.
+    $today = (new DateTime())->format('Y-m-d');
```
davialexandre: I like using `DateTime` objects, but here I think a simple `date('Y-m-d')` would be better and more readable. What do you think?

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -38,118 +38,115 @@ public static function create(&$params) {
...
   /**
-   * Search for approved Documents having expire date less than given $expireDate
-   * and create their clone.
-   * Return a count of cloned documents.
-   * 
-   * @param DateTime $expireDate
-   * @return int
+   * Creates a copy of a document with no attached files
+   *
+   * @param array $original
```
davialexandre: Please add a description to this param

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -38,118 +38,115 @@ public static function create(&$params) {
...
-   * PCHR-1365 requirements.
-   * 
-   * @param array $document
-   * @param int $expireId
-   * @return array
+   * Gets the custom field name for Documents
+   *
+   * @param string $fieldName
+   *  The human-readable name for the field, e.g. "expiry_date"
+   * @return mixed|null
```
davialexandre: Please add an empty line between the param and return sections

davialexandre: Why mixed? Shouldn't it be string|null?

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
mickadoo: I can delete it, just the IDE doesn't know that $instance is a CRM_Tasksassignments_DAO_Document so I added it to hide the warning that it was returning something unexpected. The parent class also has a return which includes `object`.

image

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
mickadoo: I'll add "or equal to"

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
mickadoo: I've found working with DateTime much more readable in the majority of cases, particularly when adding / subtracting intervals but I think in this case there's no need. Also I guess just because they introduced `DateTime` doesn't mean they're planning to deprecate other date functions.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
mickadoo: It's just expecting the same return value as `CRM_Utils_Array::value`, which is `mixed`, it was autogenerated and since it doesn't complain if I use `string` I'll change it.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/BAO/Document.php
```diff
@@ -38,118 +39,116 @@ public static function create(&$params) {
...
-      // so so we'll know that the document has been already cloned.
-      civicrm_api3('Document', 'create', array(
-        'id' => $originalId,
-        'clone_date' => (new DateTime())->format('Y-m-d'),
-      ));
+  private static function getCustomFieldName($fieldName) {
+    static $fieldNames;
+
+    if (empty($fieldNames)) {
+      $result = civicrm_api3('Document', 'getcustomfields');
```
mickadoo: @davialexandre I also made changes here since your review. I think this is simpler than those fancy array functions

davialexandre:
Approving the changes even though the tests are not passing, so it doesn't block further development on PCHR-2145. @mickadoo will work on fixing the tests and will submit a separate PR for this

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
```diff
@@ -170,4 +171,65 @@ function testCreateDocumentClone() {
...
+
+    $contact = civicrm_api3('Contact', 'create', [
+      'contact_type' => 'Individual',
+      'display_name' => 'First Contact',
+      'email' => 'test1@email3215155.com',
+    ]);
+    $details1 = 'First sample document details';
+    $details2 = 'Second sample document details';
+    $today = (new DateTime())->format('Y-m-d');
+    $tomorrow = (new DateTime('tomorrow'))->format('Y-m-d');
```
davialexandre: if you're not really using the `DateTime` instances, I think this will be simpler with the `date()` function

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/TaskTest.php
```diff
@@ -1,69 +1,75 @@
...
     $upgrader->install();
 
-    // Pick one of Task types.
-    $taskTypes = civicrm_api3('Task', 'getoptions', array(
-      'field' => 'activity_type_id',
-    ));
-    $this->_taskTypeId = array_shift($taskTypes['values']);
-  }
+    $result = civicrm_api3('OptionValue', 'create', [
+      'option_group_id' => "activity_type",
```
davialexandre: use single quotes for everything here

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/TaskTest.php
```diff
@@ -1,69 +1,75 @@
...
     $upgrader = CRM_Tasksassignments_Upgrader::instance();
     $upgrader->install();
 
-    // Pick one of Task types.
-    $taskTypes = civicrm_api3('Task', 'getoptions', array(
-      'field' => 'activity_type_id',
-    ));
-    $this->_taskTypeId = array_shift($taskTypes['values']);
-  }
+    $result = civicrm_api3('OptionValue', 'create', [
```
davialexandre: Unless you have a good reason not to, you should use the `tearDown` method to deleted anything you create on a `setUp` method.

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
```diff
@@ -242,6 +242,8 @@ function _civicrm_api3_document_create_spec(&$params) {
       'api.default' => 'user_contact_id',
   ];
 
+  $params['id']['title'] = "Document ID";
```
davialexandre: single quotes

PR #201, opened 2017-05-11
FAR-182: Empty Reminder Emails

mickadoo:
## Overview
Sending the daily reminder gathers a list of tasks and key dates that should be included in the e-mail. It's possible, however, that a contact can be selected because they're involved in a key date (their birthday, for example) but because of the logic for deciding who can see key date data they would receive an empty e-mail.

## Before
There is no check to see if there is any data to send in the email and it's possible to send empty mails.

## After
There is a check before sending if there is actually any content for the mail.

---

- [ ] Tests Pass
Unfortunately tests were fixed in PCHR-2145-updates-to-documents and this is still waiting to be merged.

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
@@ -279,6 +279,10 @@ class CRM_Tasksassignments_Reminder {
         $settings
       );
 
+      if (empty(array_filter($reminderData))) {
+        continue;
+      }
+
       $templateBodyHTML = CRM_Core_Smarty::singleton()->fetchWith('CRM/Tasksassignments/Reminder/DailyReminder.tpl', [
         'reminder' => $reminderData,
         'baseUrl' => CIVICRM_UF_BASEURL,
@@ -435,7 +439,7 @@ class CRM_Tasksassignments_Reminder {
    *
    * @param string $now
    *   Date from where key dates and appraisals should be searched, yyyy-mm-dd
-   * @param type $to
+   * @param string $to
    *   Date until where key dates and appraisals should be searched, yyyy-mm-dd
    *
    * @return string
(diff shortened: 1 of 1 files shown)


tunbola:
Looks good @mickadoo, I'll recommend you actually squash the commits because the latest commit is actually all that is needed since its an improvement over the first approach.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
```diff
@@ -427,25 +431,25 @@ private static function _buildTaskAssigneeCreatorQuery($to) {
...
    *   dates and appraisals
    */
   private static function _buildAdminsKeyDatesAndAppraisalsQuery($now, $to) {
     $adminContacts = self::_getAdminContactIds();
     $keyDatesContacts = CRM_Tasksassignments_KeyDates::getContactIds($now, $to);
-    $appraisalsContacts = self::$_relatedExtensions['appraisals'] 
-      ? CRM_Appraisals_Reminder::getContactIds($now, $to) 
+    $appraisalsContacts = self::$_relatedExtensions['appraisals']
+      ? CRM_Appraisals_Reminder::getContactIds($now, $to)
       : array();
```
tunbola: Please use short array syntax for the empty array here.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
```diff
@@ -427,25 +431,25 @@ private static function _buildTaskAssigneeCreatorQuery($to) {
...
+   *
    *   - Users with administrator or civihr_admin roles
    *   - Contacts with key dates in given time period
    *   - Appraisals due in the given timeframe
-   * 
+   *
    * @param string $now
-   *   Date from where key dates and appraisals should be searched, yyyy-mm-dd 
+   *   Date from where key dates and appraisals should be searched, yyyy-mm-dd
    * @param type $to
```
tunbola: I think we need to fix this param type.

PR #204, opened 2017-05-22
PCHR-1928: Hide application case type from view

mickadoo:
## Overview
The application case type is used by hrrecruitment extension, but it is not an assignment in the definition that T&A uses. Users can select it, but it is not very usable since most of the tasks for it will appear as disabled in the modal.

## Before
The Application case type appears in both the Assignment edit page and the dropdown when selecting an assignment type.

## After
The Application case type is hidden.

- [ ] Tests Pass
Tests are still broken until #196 is merged.


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Wrapper/StripInvalidCaseTypesWrapper.php
```diff
@@ -0,0 +1,28 @@
...
+   */
+  public function fromApiInput($apiRequest) {
+    $apiRequest['params']['name'] = ['NOT IN' => $this->invalidTypeNames];
+
+    return $apiRequest;
+  }
+
+  /**
+   * @param array $apiRequest
+   * @param array $result
```
omarabuhussein: space between @param and @return

omarabuhussein:
oh !! actually can you provide before and after screenshots , or you don't do this anymore for civihr ?

mickadoo:
Closing since the decision for PCHR-1928 was to remove the override file (https://github.com/civicrm/civihr/pull/1871) and keep other case types for now.

PR #205, opened 2017-05-26
PCHR-2144: Updates to SSP Documents  widget 

bmnepali:
## Overview
The markup (`/views/modal/document.html`) in the New Document Modal was not inline with bootstrap standards, which is breaking the modal even adding bootstrap styles. So, refactoring the markup is required to match with bootstrap standard, keeping the modal in T&A intact.

Also, since the `DocumentCtrl` is not used the SSP Document section, it is never initialized, which formats the date fields, target contact, assignee and remind me status. We need them as well in Document modal. So, those scripts are added in `/js/src/tasks-assignments/controllers/modal/modal-document.js` file.

## Before
1. Before the markup refactoring to match bootstrap standards.
screen shot 2017-05-29 at 5 01 58 pm

## After
Step 1. Modal Window in SSP, once the markup is refactored.
screen shot 2017-05-25 at 6 17 22 pm

Step 2. Final Modal window, once the custom styles are added and hidden fields for the staff role
screen shot 2017-06-01 at 1 19 20 pm

## Technical Details
1. "staff' role is used to hide the respective fields in the document modal markup in `uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html`. The role is passed to the Modal Document Controller in `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js`  and stored in scope. The role is passed  main controller in `js/src/tasks-assignments/controllers/main.js`  as following (role to be 'admin' by default)
```js
  $rootScope.modalDocument = function (data, e) {
        ...
        var modalInstance = $modal.open({
          ...
          resolve: {
            role: function () {
              return ['admin'];
            },
            ...
          }
        });
```
2. The `uib-tabset` is is replace by the regular markup for `nav-tabs` from bootstrap. The main reason being `uib-tabset` being angular directive, I could not hide the tabs only leaving the tab content still visible, which was easily possible in regular tab markup.

3. The Documents modal was made originally to work only on the admin, so it assumed the user would always have admin permissions. Now that civihr_staff users can also access it, some permissions changes were necessary:
- In order to upload, see and remove files from  a document, the user needs access to some callback pages. Originally, the permission required to this page was `access Tasks and Assignments`. The problem is that this same permission is also used to some admin only pages, meaning civihr_staff can't have it. A new `access Tasks and Assignments Files` was created only for the file related pages, so both admins and civihr_staff can have it without risking exposing other admin pages to staff.
- The modal makes API calls to the CaseType.get API, which, by default, required the "access all cases and activities" permission. Civihr_staff users don't have this permission, so, in order to avoid access denied errors, `hook_civicrm_alterAPIPermissions` was used to tell civi to no check permissions for this API. Note that this is not the best and most secure way to approach the problem, but it's the way it was already done for other APIs in T&A. We should definitely improve this later
- The modal also makes calls to the Document.get API, which internally calls the Activity.get API, which also requires its own special permissions. The same approach as above was used for this API.

- Pre-commit linter was displaying warning about the naming convention for `HR_settings` which is a factory service defined in `reqangular` extension. I am skipping this lint warning as this needs to change its name across multiple extensions.

### Update:
1. Added styles to change the background of "Remove" button to white on focus, in list of attached documents.
2. Amend title markup of Document modal to display the document type in the Modal, when opened form SSP as staff or manager. But keep the title as "New Document" for admin. Added test cases to support additional js change.
3. Make changes to modal-document.js, on saving form, 
-> display the confirmation modal, if there are changes in list of documents have changed.
-> Make the status of document to be "Awaiting Approval" when saved by staff or manager from SSP
-> Make the status of document to be "Awaiting Upload" when saved by staff or manager from SSP, but if no documents are uploaded.

- [x] Tests Pass


mickadoo:
@bmnepali Since PCHR-2145 is merged to staging now can you change the target branch for this and delete the 2145 branch please?

bmnepali:
@mickadoo Will do. Thanks!

dhavalsharma:
modal-document.js file name looks generic, can it be more specific like tna-document or something like tht

bmnepali:
@dhavalsharma Ignored, your first comment as there are more modals and they look good as they are now. And updated you second comment.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -151,11 +151,17 @@ function tasksassignments_civicrm_alterSettingsFolders(&$metaDataFolders = NULL)
   _tasksassignments_civix_civicrm_alterSettingsFolders($metaDataFolders);
 }
 
-function tasksassignments_civicrm_alterAPIPermissions($entity, $action, &$params, &$permissions)
-{
-    $permissions['contact']['get'] = array();
-    $permissions['contact']['getquick'] = array();
-  if ($entity == 'document' || $entity == 'task' || $entity == 'assignment') {
+function tasksassignments_civicrm_alterAPIPermissions($entity, $action, &$params, &$permissions) {
+  $permissions['contact']['get'] = array();
```
mickadoo: even though it's just whitespace that changed on these lines it might be nice to use the `[]` array format for this function

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -151,11 +151,17 @@ function tasksassignments_civicrm_alterSettingsFolders(&$metaDataFolders = NULL)
...
+  $permissions['contact']['getquick'] = array();
+  $entitiesToAvoidPermissions = [
```
mickadoo: Does this mean that any API call to any of these entities will be allowed from any user?

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -248,6 +248,7 @@ function tasksassignments_civicrm_permission(&$permissions) {
   $permissions += array(
     'delete Tasks and Documents' => $prefix . ts('delete Tasks and Documents'),
     'access Tasks and Assignments' => $prefix . ts('access Tasks and Assignments'),
+    'access Tasks and Assignments Files' => $prefix . ts('access Tasks and Assignments Files'),
```
mickadoo: Just a question, I added a new permission in civihr for another ticket and added it to a default permissions file. Is this different here?

mickadoo: Never mind, I found it in the other PR on the employee portal

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
davialexandre: Unfortunately, yes. As I mentioned on the PR description, this is not a secure way to do this. I'm just following the existing approach in the extension. We need to fix a lot of things regarding T&A and API permissions, but that would be out of the scope of this ticket. I'm going to flag this to Mike and hopefully we'll have another ticket to handle this properly

mickadoo: > We should definitely improve this later
Definitely agree there, this is a big security hole.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -152,15 +152,17 @@ function tasksassignments_civicrm_alterSettingsFolders(&$metaDataFolders = NULL)
 }
 
 function tasksassignments_civicrm_alterAPIPermissions($entity, $action, &$params, &$permissions) {
-  $permissions['contact']['get'] = array();
-  $permissions['contact']['getquick'] = array();
+  $permissions['contact']['get'] = [];
+  $permissions['contact']['getquick'] = [];1
```
mickadoo: eh, I just noticed this now myself..

bmnepali: will fix it!

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js
```diff
@@ -1,3 +1,5 @@
+/* globals define, jasmine, describe, beforeEach, it, expect, spyOn */
```
AkA84: https://github.com/feross/standard#what-about-mocha-jasmine-qunit-etc
```js
/* eslint-env jasmine */
```

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js
```diff
@@ -7,20 +9,20 @@ define([
...
+    var $controller, $rootScope, $scope, hrSettings, data, role, files, sampleAssignee;
 
     beforeEach(module('civitasks.appDashboard'));
     beforeEach(inject(function (_$controller_, _$rootScope_) {
       $controller = _$controller_;
       $rootScope = _$rootScope_;
       $scope = $rootScope.$new();
 
-      HR_settings = { DATE_FORMAT: 'DD/MM/YYYY' };
+      hrSettings = { DATE_FORMAT: 'DD/MM/YYYY' };
```
AkA84: `HRSettings`

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/main.js
```diff
@@ -19,6 +19,9 @@ define([
           templateUrl: config.path.TPL + 'modal/document.html?v=3',
           controller: 'ModalDocumentCtrl',
           resolve: {
+            role: function () {
+              return ['admin'];
```
AkA84: why is this needed to be an array? Why not simply a string?

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js
```diff
@@ -25,6 +25,7 @@ define([
 
       data = {};
       files = {};
+      role = 'admin';
```
AkA84: are you testing the `role` property in any way?

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -9,7 +9,7 @@ <h4 class="modal-title">New Document</h4>
   <div class="modal-body" nv-file-drop uploader="uploader">
     <div class="container">
       <!-- Search and select by name and email -->
-      <div class="row">
+      <div class="row" ng-hide="role === 'staff'">
```
AkA84: you have this logic duplicated everywhere in the view. Encapsulate it in an helper method that you publish on the scope, something like
```html
<div class="row" ng-hide="isRole('staff')">
```

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -39,8 +39,12 @@ <h4 class="modal-title">New Document</h4>
       </div>
     </div>
     <div class="container">
-      <uib-tabset>
```
AkA84: Why are you getting rid of uib-tabset? I don't think you mentioned it in the technical details

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js
bmnepali: Not testing! but the while instantiating the controller it expects me to pass role.

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js
bmnepali: ok! will do

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
bmnepali: right! will update technical steps.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -58,6 +58,10 @@ define([
         $scope.remindMeMessage = 'If you check this box CiviHR will “remind” you that this document needs to be reviewed. CiviHR will do this by creating a copy of  the document with the  status of awaiting upload a number of days or months before the documentexpires. You can set the date to create the copy. The copy will have the same document  types and set the assignee to be the same assignee as for this original version of the document. You will then see it in your documents list and be able to action renewing the document.';
       })();
 
+      $scope.isRole = function (role) {
```
AkA84: document the function

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/main.js
```diff
@@ -37,10 +37,6 @@ define([
 
         modalInstance.result.then(function (results) {
           $scope.$broadcast('documentFormSuccess', results, data);
-
-          if (results.open) {
```
AkA84: why are you removing this?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -249,7 +253,6 @@ define([
           $scope.document.id = result.document.id;
           $scope.document.case_id = result.document.case_id;
           $scope.document.file_count = $scope.files.length + uploader.queue.length;
-          $scope.document.open = $scope.openNew;
```
AkA84: why are you removing this?

bmnepali:
```js
$scope.document.open = $scope.openNew;
```
This says weather we want to open new modal or not. Since we have removed "Save and New" button as per the wireframe, we won't be needing it any more.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/main.js
bmnepali: The above function is needed to open a document, if ` $scope.document.open` was set true. But, Since we have removed "Save and New" button as per the wireframe, we wont be needing to  pen new modal once the document is saved. So, we won't be needing it any more.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
bmnepali: will do!

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/main.js
bmnepali: Fixed it!

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -6,65 +8,79 @@ define([
...
+        };
 
-      $scope.remindMeMessage = "If you check this box CiviHR will “remind” you \
-        that this document needs to be reviewed. CiviHR will do this by creating \
-        a copy of  the document with the  status of awaiting upload a number of days \
-        or months before the documentexpires. You can set the date to create the copy. \
-        The copy will have the same document  types and set the assignee to be the same \
-        assignee as for this original version of the document. You will then see it in \
-        your documents list and be able to action renewing the document.";
+        $scope.remindMeMessage = 'If you check this box CiviHR will “remind” you that this document needs to be reviewed. CiviHR will do this by creating a copy of  the document with the  status of awaiting upload a number of days or months before the documentexpires. You can set the date to create the copy. The copy will have the same document  types and set the assignee to be the same assignee as for this original version of the document. You will then see it in your documents list and be able to action renewing the document.';
```
mickadoo: documentexpires should be two words

bmnepali: good catch! will fix it.

uk.co.compucorp.civicrm.tasksassignments/js/test/mocks/document.js
```diff
@@ -0,0 +1,27 @@
+/* globals define */
```
AkA84: /* eslint-env amd */

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js
```diff
@@ -1,20 +1,23 @@
+/* globals define, inject, CRM */
 /* eslint-env jasmine */
```
AkA84:
```js
/* globals CRM, inject */
/* eslint-env amd, jasmine */

File: uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js [hunk 17]
```diff
@@ -129,6 +145,31 @@ define([
       });
     });
 
+    describe('$scope.role', function () {
+      beforeEach(function () {
+        initController();
+      });
+
+      it('defaults to as admin', function () {
```
AkA84: all the initialization logic of the controller should be under the same section. You already have a `describe init()` block, put this there
```js
describe('init()', function () {
  beforeEach(function () {
    // ...
    initController();
  });

  it('sets the role as admin by default', function () {
  });

  it('correctly formats date time in document', function () {
  });
});

File: uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js [hunk 18]
```diff
@@ -129,6 +145,31 @@ define([
...
+      });
+    });
+
+    describe('isRole()', function () {
+      beforeEach(function () {
+        initController();
+        $scope.role = 'staff';
+      });
+
+      it('returns truthy value for correct username', function () {
```
AkA84: this test title doesn't make much sense related to what you are testing (what does the username has anything to do with it?)

```js
it('checks if the given role is the current one', function () {
  expect($scope.isRole('staff')).toBe(true);
  expect($scope.isRole('admin')).toBe(false);
});

File: uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js [hunk 19]
```diff
@@ -129,6 +145,31 @@ define([
...
+      });
+
+      it('defaults to as admin', function () {
+        expect($scope.role).toBe('admin');
+      });
+    });
+
+    describe('isRole()', function () {
+      beforeEach(function () {
+        initController();
```
AkA84: amend `initController` so you can pass default values to the scope
```js
initController({ role: 'staff' });

// ...
function initController(scopeValues) {
  $controller('ModalDocumentCtrl', {
    $scope: _.assign($scope, scopeValues),
  }
}

File: uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js (hunk 17 above)
Comment (bmnepali) [inline uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js]: will do!

File: uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js (hunk 18 above)
Comment (bmnepali) [inline uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js]: Right! It should be user role rather that 'username'. Will fix it.

File: uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js (hunk 19 above)
Comment (bmnepali) [inline uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js]: Impleented following code 
```js
// ...
function initController(scopeValues) {
  $controller('ModalDocumentCtrl', {
    $scope: _.assign($scope, scopeValues),
  }
}
```

But, here, `$scope.role` even passed as `initController({ role: 'staff' });`, it will get overridden by the value of `role`. As the  `$scope.role` is dependent on `role`. So, in this case, we need to do `$scope.role = 'staff';` after controller is initialized.
```js
describe('isRole()', function () {
      beforeEach(function () {
        initController();
        $scope.role = 'staff';
      });

      it('checks if the given role is the current one', function () {
        expect($scope.isRole('staff')).toBe(true);
        expect($scope.isRole('admin')).toBe(false);
      });
    });
```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -260,6 +260,7 @@ define([
           $scope.document.file_count = $scope.files.length + uploader.queue.length;
 
           AssignmentService.updateTab();
+          $rootScope.$broadcast('document-saved');
```
AkA84: leave an empty line
```js
AssignmentService.updateTab();

$rootScope.$broadcast('document-saved');
$scope.$broadcast('ta-spinner-hide');
```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -67,6 +67,21 @@ define([
...
+       * @param  {integer} documentTypeId
+       * @return {string}
+       */
+      $scope.getDocumentType = function (documentTypeId) {
+        var documentTypes = $rootScope.cache.documentType.arr;
+        var documents = documentTypes.map(function (document) {
+          return document.key;
+        });
+
+        return documentTypeId && documentTypes[documents.indexOf(documentTypeId)].value;
```
AkA84: isn't this equivalent to
```js
/**
 *
 */
$scope.getDocumentType = function (documentTypeId) {
  return _.find($rootScope.cache.documentType.arr, function (documentType) {
    return documentType.key === documentTypeId;
  }).value;
};
```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -216,6 +216,7 @@ define([
 
         doc.activity_date_time = $scope.parseDate(doc.activity_date_time) || new Date();
         doc.expire_date = $scope.parseDate(doc.expire_date);
+        doc.status_id = (!$scope.isRole('admin')) ? '2' : $scope.document.status_id; // 2 => 'Awaiting Approval'
```
AkA84: 1. `(!$scope.isRole('admin'))` the outer parenthesis are not necessary `!$scope.isRole('admin')`
2. Why is the status id hardcoded?
3. So if i'm staff or manager the id is always going to be "Awaiting approval"?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -261,7 +262,7 @@ define([
           } else if (result.files.length && !!result.files[0].values[0].result) {
             DocumentService.save({
               id: result.document.id,
-              status_id: '3' // 3 => 'approved'
+              status_id: ($scope.isRole('admin')) ? '3' : '1' // 3 => 'approved', 1 => 'awaiting upload'
```
AkA84: 1. `($scope.isRole('admin'))` the outer parenthesis are not necessary `$scope.isRole('admin')`
2. Why is the status id hardcoded?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -173,7 +173,7 @@ define([
       };
 
       $scope.cancel = function () {
-        if ($scope.documentForm.$pristine) {
+        if ($scope.documentForm.$pristine && angular.equals(files, $scope.files) && !$scope.uploader.queue.length) {
```
AkA84: can you tell me a bit more what's going on here?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
bmnepali: > Why is the status id hardcoded?
As the status id of `Awaiting Approval` is always `2`, not not going to change, so, it is hard coded at the moment.

> So if i'm staff or manager the id is always going to be "Awaiting approval"?
Not always, if the document is updated with attachment the the status will be `Awaiting Approval`. But if, attachment are not attached the status of the document will be `Awaiting Upload`.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
bmnepali: It is! will use it.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
bmnepali: Currently,  the confirmation popup is only displayed if not fields data (except file attachment changes) in fields are changed, and when he tries to cancel the form. So, i have added two more conditions:
1. `angular.equals(files, $scope.files)` which is to check if the files  have been changed
2.`!$scope.uploader.queue.length` which is to check if there are files in the  uploader queue.
I am displaying the confirmation, also for when there is change in files and when there are files in the uploader queue and if user tries to cancel the form.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
bmnepali: > Why is the status id hardcoded?

The status id `2` which  corresponds to `Awaiting Approval` is hardcoded, as this is constant and never changes.

>So, for staff and manager the status is always going to be `Awaiting Approval`?

Not really, the status will be `Awaiting Approval` if the document is updated with attachments attached, otherwise the status will be `Awaiting Upload`

bmnepali: > Why is the status id hardcoded?

The status id `2` which  corresponds to `Awaiting Approval` is hardcoded, as this is constant and never changes.

>So, for staff and manager the status is always going to be `Awaiting Approval`?

Not really, the status will be `Awaiting Approval` if the document is updated with attachments attached, otherwise the status will be `Awaiting Upload`

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -1,4 +1,5 @@
-/* global define, CRM */
+/* global CRM, _ */
```
AkA84: `_` can't be global, inject the `common/lodash` requirejs module

bmnepali: Fixed!

PR #207, opened 2017-06-02
PCHR-2146: Updates for document management widget civihr_manager

bmnepali:
## Overview
All Fields for New Document Modal are visible  to admin in T&A. and since we are reusing the Modal in SSP, respective fields, as shown in the wireframe, needs to be displayed and hidden for the civihr_manager user.

## Before
screen shot 2017-06-02 at 5 10 40 pm
I believe, the ellipses in the right of each document, has a option (dropdown) to display a modal in Documents manager section, but since i can't open it, i can't see the before screenshot of the Document Modal.

## After
screen shot 2017-06-01 at 9 09 31 pm

## Technical Details
1. Added `ng-hide`, `ng-show` and `ng-if` respectively, in the html markup to appropriately display and hide the fields for the `civihr_manager` user.

2. Currently, in T&A has two tabs, "Details" and "Assignments". The Details tab contains the fields like target contact, document type, from date, expiry date, status and remind me and Assignments tab contains the fields like due date, assignee and link to assignment. 
Now, when the Modal is opened from SSP as civihr_manager, only one  tab (view) should be displayed rather than tabs. And since the fields to show for Manager are target contact, documents type, from date, expiry date, status and due date. Except "due date" field which is in Assignments tab, all other are available in details tab content. 
So,  to make the modal to display the fields for manager in single view (details tab content), i have duplicated the due date to be in details tab,  this is removed from DOM for other users due to the use of `ng-if` angular directive.
```html
<!-- Display only for manager role and hidden removed from DOM for other uasers -->
            <div class="col-xs-6" ng-if="isRole('manager')">
              <div class="form-group">
                <label for="{{prefix}}document-due" class="text-nowrap control-label required-field-indicator"> Due Date: </label>
                <div class="input-group" ng-class="{'has-error': documentForm.dueDate.$invalid}">
                  <input name="dueDate" type="text" id="{{prefix}}document-due" class="form-control" ng-model="document.activity_date_time" placeholder="dd/mm/yyyy" uib-datepicker-popup ng-required="true" is-open="dpOpened.due" ng-click="dpOpen($event, 'due')">
                  <span class="input-group-addon fa fa-calendar"></span>
                </div>
              </div>
            </div>
```
### Update 1:
1. Modify `$scope.getDocumentsType()` function to check if the document type is available or not.
```js
$scope.getDocumentType = function (documentTypeId) {
        var documentTypes = _.find($rootScope.cache.documentType.arr, function (documentType) {
            return documentType.key === documentTypeId;
          });

        return documentTypes && documentTypes.value;
      };
```
2. Added `Select Document Type` text in document title in modal, if  the document type in form   is cleared.
```html
<span ng-switch-when="false">{{getDocumentType(document.activity_type_id) || 'Select Document Type'}}</span>
```

### Update 2:
1. In T&A, `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js` has been updated to get the contact for the given documents list and then fetch the contacts and assignments with the addition of following functions:

i. `collectContactIds` collects the ids from the list of documents.
```js
function collectContactIds (documentList) {
...
};
```

ii. `collectAssignmentIds` collects the ids from the list of documents.
```js
function collectAssignmentIds (documentList) {
...
};
```

iii. `cacheContactsAndAssignments` catches the contacts and assignments.
```js
cacheContactsAndAssignments: function (documentList, options) {
...
};
```
2. Using the updated document service, caching the contacts and assignments in T&A in file  `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/document-list.js` as follows:

```js
     this.init = function(){
          DocumentService.cacheContactsAndAssignments(documentList).then(function() {
             ....
           });
           ...
      };
```

### Comments:
currently i have skipped these two linter issues, the code seems to be as written and could nt  find the solution to disable `no-mixed-operators`, in `js/src/tasks-assignments/controllers/document-list.js` file.
```js
  262:40  error  Unexpected mix of '||' and '&&'  no-mixed-operators
  263:71  error  Unexpected mix of '||' and '&&'  no-mixed-operators
```


uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -23,7 +23,7 @@ <h4 class="modal-title">New Document</h4>
         </div>
       </div>
       <!-- Document Type -->
-      <div class="row" ng-hide="isRole('staff')">
+      <div class="row" ng-hide="isRole('staff')" ng-show="!isRole('manager')">
```
deb1990: Do not include `ng-hide`and `ng-show` in the same markup, you can just add your new condition in `ng-hide`with `||` or `&&` as applicable

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -118,7 +118,7 @@ <h4 class="modal-title">New Document</h4>
               </div>
             </div>
             <!-- Document Status -->
-            <div class="col-xs-12 col-sm-6" ng-hide="isRole('staff')">
+            <div class="col-xs-12 col-sm-6" ng-hide="isRole('staff')"  ng-show="isRole('manager')">
```
deb1990: Same as above

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -141,22 +141,33 @@ <h4 class="modal-title">New Document</h4>
                 </div>
               </div>
             </div>
+
+            <!-- Display only for manager role and hidden removed from DOM for other uasers -->
+            <div class="col-xs-6" ng-if="isRole('manager')">
+              <div class="form-group">
+                <label for="{{prefix}}document-due" class="text-nowrap control-label required-field-indicator"> Due Date: </label>
+                <div class="input-group" ng-class="{'has-error': documentForm.dueDate.$invalid}">
+                  <input name="dueDate" type="text" id="{{prefix}}document-due" class="form-control" ng-model="document.activity_date_time" placeholder="dd/mm/yyyy" uib-datepicker-popup ng-required="true" is-open="dpOpened.due" ng-click="dpOpen($event, 'due')">
```
deb1990: This line is too big, break this down
```html
<input
  a="..."
  b="..."
```

or 

```html
<input
  a="..."  b="..."
  c="..."  d="..."
```

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -141,22 +141,33 @@ <h4 class="modal-title">New Document</h4>
...
-              <label for="{{prefix}}document-due" class="col-xs-4 text-nowrap control-label required-field-indicator"> Due Date: </label>
-              <div class="col-xs-6 input-group" ng-class="{'has-error': documentForm.dueDate.$invalid}">
-                <input name="dueDate" type="text" id="{{prefix}}document-due" class="form-control" ng-model="document.activity_date_time" placeholder="dd/mm/yyyy" uib-datepicker-popup ng-required="true" is-open="dpOpened.due" ng-click="dpOpen($event, 'due')">
-                <span class="input-group-addon fa fa-calendar"></span>
+            </br>
+            <div class="col-xs-10 col-sm-12">
+              <div class="form-group">
+                <label for="{{prefix}}document-due" class="col-xs-4 text-nowrap control-label required-field-indicator"> Due Date: </label>
+                <div class="col-xs-6 input-group" ng-class="{'has-error': documentForm.dueDate.$invalid}">
+                  <input name="dueDate" type="text" id="{{prefix}}document-due" class="form-control" ng-model="document.activity_date_time" placeholder="dd/mm/yyyy" uib-datepicker-popup ng-required="true" is-open="dpOpened.due" ng-click="dpOpen($event, 'due')">
```
deb1990: Break the line as I mentioned above

deb1990:
@bmnepali You say 

> There was not a option to display a modal in Documents manager section

Could you point me which part of code is providing the option now?

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -141,22 +141,33 @@ <h4 class="modal-title">New Document</h4>
                 </div>
               </div>
             </div>
+
+            <!-- Display only for manager role and hidden removed from DOM for other uasers -->
+            <div class="col-xs-6" ng-if="isRole('manager')">
```
deb1990: Why sometimes you are using `ng-hide/show`and sometimes `ng-if`?

deb1990:
@bmnepali I dont understand "Due date field is duplicated, just to make the modal to display the fields for manager in single view" which single view are you refering to?

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
bmnepali: ok will do!

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
bmnepali: 1. `hg-hide/ng-show`  make the element invisible, but still keep in DOM.
2. `ng-if` to remove the element from the DOM.

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
bmnepali: ok will do!

bmnepali:
@deb1990 I have updated the PR description to address your comments. Please check.

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -141,22 +141,41 @@ <h4 class="modal-title">New Document</h4>
                 </div>
               </div>
             </div>
+
+            <!-- Display only for manager role and hidden removed from DOM for other uasers -->
```
AkA84: typo

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -141,22 +141,41 @@ <h4 class="modal-title">New Document</h4>
                 </div>
               </div>
             </div>
+
```
AkA84: remove the empty line

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
AkA84: but is there consistency here?

for example you have
```html
<div class="tab-pane" id="assignmentsTab" ng-hide="isRole('staff') || isRole('manager')">
```
why ng-hide in this case?

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
bmnepali: will fix

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
bmnepali: will fix

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
bmnepali: Well, thanks for catching! Since we wont be needing the Assignments tab, for both users (staff and manager), it's more appropriate not to keep it in the DOM. So, this
```html
<div class="tab-pane" id="assignmentsTab" ng-hide="isRole('staff') || isRole('manager')">
```
should be 
```html
<div class="tab-pane" id="assignmentsTab" ng-if="!isRole('staff') && !isRole('manager')">
```

bmnepali:
@AkA84 Thanks for catching!  It seems that we won't be needing the assignments tab for both staff and manager. So it's more appropriate to use `ng-if`  which will remove the tab content form the DOM. Where as `ng-hide` hide the contents but will still keep it in the DOM. So, this
```html
<div class="tab-pane" id="assignmentsTab" ng-hide="isRole('staff') || isRole('manager')">
```
should be
```html
<div class="tab-pane" id="assignmentsTab" ng-if="!isRole('staff') && !isRole('manager')">
```
Also,  here `ng-if` is used so the element is removed rather than hidden form the DOM for due date field.
```html
 <div class="col-xs-6" ng-if="isRole('manager')">
```

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
```diff
@@ -6,7 +6,7 @@
     </button>
     <h4 class="modal-title" ng-switch="isRole('admin')">
       <span ng-switch-when="true">New Document</span>
-      <span ng-switch-when="false">{{getDocumentType(document.activity_type_id)}}</span>
+      <span ng-switch-when="false">{{getDocumentType(document.activity_type_id) || 'Select Document Type'}}</span>
```
AkA84: so the logic here is:

If the modal is being opened by an admin, then *always* show as a title "New Document", otherwise if it's being opened by a staff or a manager, show as a title either the document type or a message asking to select the type

but why does that admin always see "New Document"? What if the admin open an already existing document?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -77,9 +77,11 @@ define([
        * @return {string}
        */
       $scope.getDocumentType = function (documentTypeId) {
-        return _.find($rootScope.cache.documentType.arr, function (documentType) {
+        var documentTypes = _.find($rootScope.cache.documentType.arr, function (documentType) {
```
AkA84: `documentType`, it's just one

uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html
bmnepali: For now, in  the title will remain "New Document" in edit mode. Not sure if we need to change to a different title for edit mode. But yes, doesn't fit for edit mode.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
bmnepali: will fix it.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
```diff
@@ -17,10 +17,40 @@ define([
     });
   }]);
 
-  services.factory('DocumentService', ['Document', '$q', 'config', 'UtilsService', '$log',
-    function (Document, $q, config, UtilsService, $log) {
+  services.factory('DocumentService', ['Document', '$q', 'config', 'settings', 'UtilsService', 'ContactService', 'AssignmentService', '$log',
+    function (Document, $q, config, settings, UtilsService, ContactService, AssignmentService, $log) {
       $log.debug('Service: DocumentService');
 
+      function collectContactIds (documentList) {
```
AkA84: document the function

AkA84: move private functions below the `return`, hide the implementation details at the bottom

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
```diff
@@ -17,10 +17,40 @@ define([
...
+
+          if (document.target_contact_id && document.target_contact_id.length) {
+            contactIds.push(document.target_contact_id[0]);
+          }
+        });
+
+        return contactIds;
+      }
+
+      function collectAssignmentIds (documentList) {
```
AkA84: document the function

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
AkA84: the param could be simply called `documents` instead of `documentList`

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
```diff
@@ -17,10 +17,40 @@ define([
...
+          if (document.assignee_contact_id && document.assignee_contact_id.length) {
+            contactIds.push(document.assignee_contact_id[0]);
+          }
+
+          if (document.target_contact_id && document.target_contact_id.length) {
+            contactIds.push(document.target_contact_id[0]);
+          }
+        });
+
+        return contactIds;
```
AkA84:
```js
function collectContactIds (documens) {
  return _(documents).map(function (document) {
    var foo = [];
     // push all the different contacts in foo

     return foo;
  }).flatten().value();
```
see if it's the same as the original solution

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
AkA84: `documents` instead of `documentList`

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
```diff
@@ -17,10 +17,40 @@ define([
...
+      function collectAssignmentIds (documentList) {
+        var assignmentIds = [];
+
+        angular.forEach(documentList, function (document) {
+          if (document.case_id) {
+            assignmentIds.push(document.case_id);
+          }
+        });
+
+        return assignmentIds;
```
AkA84:
```js
function collectAssignmentIds (documents) {
  return documents.filter(function (document) {
    return !!document.case_id;
  })
  .map(function () {
    return document.case_id;
  });
}

File: uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js [hunk 14]
```diff
@@ -242,6 +272,41 @@ define([
           });
 
           return deferred.promise;
+        },
+        CacheContactsAndAssignments: function (documentList, options) {
```
AkA84: document your code

AkA84: whenever you write a new method, always think first how people would use it. Even better, how *you* would like the method to work if you were to just be the "consumer" of the method, and not the author. Right now it just looks like this
```js
DocumentService.CacheContactsAndAssignments(myDocuments, ['contacts', 'assignments']);
```
If you put yourself in the shoes of a developer that would use your method, several questions would pop up

1. Why is the method beginning with a capital letter? This looks like a constructor method!
2. The method is called "cache contacts and assignments", why do I need to specify, in the second param, that i want to cache contacts and assignments? It's already there in the name, i would expect the method to do it automatically! What would be natural to expect is that by default it caches both contacts and assignments
    ```js
    DocumentService.CacheContactsAndAssignments(myDocuments);
   ```
    but that, if i wanted, i could say explicitly that i want to cache only one of the two
    ```js
    DocumentService.CacheContactsAndAssignments(myDocuments, 'contacts');
   ```
3. How do i know when the method is done? It's performing an async operation, but it's not returning a promise or anything! I would expect to be able to do something like this
    ```js
    DocumentService.CacheContactsAndAssignments(myDocuments).then(function () {
      // they are cached, moving on...
    });
   ```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
```diff
@@ -242,6 +272,41 @@ define([
...
+          var contactIds = [];
+          var assignmentIds = [];
+
+          angular.forEach(options, function (type) {
```
AkA84: `forEach` on arrays is available natively in ECMAScript 5
```js
options.forEach(function (option) {

}
```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
```diff
@@ -242,6 +272,41 @@ define([
...
+            switch (type) {
+              case 'contacts':
+                contactIds = collectContactIds(documentList);
+
+                if (config.CONTACT_ID) {
+                  contactIds.push(config.CONTACT_ID);
```
AkA84:
```js
config.CONTACT_ID && contactIds.push(config.CONTACT_ID);
```

AkA84: wasn't there also another one, like `config.LOGGED_IN` or something like that

uk.co.compucorp.civicrm.tasksassignments/js/test/services/document_test.js
```diff
@@ -0,0 +1,77 @@
...
+    beforeEach(module('civitasks.appDashboard'));
+    beforeEach(inject(function (_AssignmentService_, _ContactService_, _DocumentService_, _$q_) {
+      AssignmentService = _AssignmentService_;
+      ContactService = _ContactService_;
+      DocumentService = _DocumentService_;
+      $q = _$q_;
+      deferred = $q.defer();
+    }));
+
+    describe('collectAssignmentIds()', function () {
```
AkA84: Unit tests should cover the *public interface* of a service. "collectAssignmentIds()" is a private internal function, it doesn't make sense to make it the subject of the test. This is how the test suite for `DocumentService` should look like
```js
describe('DocumentService', function () {

  describe('CacheContactsAndAssignments', function () {

    it('caches the contacts and assignments of the given documents', function () {

    });

    it('can cache either contacts or assignments exclusively', function () {

    });

    it('returns a promise', function () {

    });
  });
});
```
that's it. Notice that the test is written from the point of view of the "consumer" of the service! It describes the expectation of how the service should behave when used, it doesn't care about the internal functioning

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/document-list.js
```diff
@@ -17,22 +17,8 @@ define([
...
-                        ContactService.updateCache(data);
-                    });
-                }
-
-                if (assignmentIds && assignmentIds.length && settings.extEnabled.assignments) {
-                    AssignmentService.get({'IN': assignmentIds}).then(function(data){
-                        AssignmentService.updateCache(data);
-                    });
-                }
+                DocumentService.CacheContactsAndAssignments(documentList, ['contacts', 'assignments']);
```
AkA84: is there a 4-spaces indentation here? If yes, it needs to be fixed

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document-list_test.js
```diff
@@ -0,0 +1,41 @@
...
+      deferred = $q.defer();
+    }));
+
+    describe('init()', function () {
+      beforeEach(function () {
+        spyOn(DocumentService, 'cacheContactsAndAssignments').and.returnValue(deferred.promise);
+        initController();
+      });
+
+      it('calls document service to cache contacts and assigments', function () {
```
AkA84: nice (thumbsup)

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
bmnepali: Right! i will try to optimize it!

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/document.js
bmnepali: Yes there was. And as we saw to call to fetch contact details of currently logged in user and caching it, is already done in `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/modules/run.js`. there is no need to fetch the contact details and cache again. So, removed from here.
```js
    ContactService.get({'IN': [config.LOGGED_IN_CONTACT_ID]}).then(function (data) {
        ContactService.updateCache(data);
      });
```

PR #222, opened 2017-06-20
PCHR-2317: Remove dropdown bar for multiple select in sidebar in T&A

bmnepali:
## Overview
The multiple select of document status in T&A is broken displaying the dropdown bar in the field. This PR removes the  dropdown bar specific to the field.

## Before
screen shot 2017-06-20 at 1 47 24 pm

## After
screen shot 2017-06-20 at 4 34 55 pm

## Technical Details
1. Added `no-dropdown` attribute in ui-select element in `uk.co.compucorp.civicrm.tasksassignments/views/dashboard/documents.html`
```html
  <ui-select ... no-dropdown>
```

2. Now add style to hide the dropdown bar in `uk.co.compucorp.civicrm.tasksassignments/scss/civitasks/modules/_civihr-ui-select.scss` file.
```css
// Do not display dropdown for ui-select-multiple with attribute no-dropdown
[no-dropdown].ui-select-multiple {
  &:after {
    display: none;
  }
}
```

### Update:
1. The modal Controller is being reused in SSP as well. Since the SSP does not have the notification as in T&A, the `var notification = CRM.alert(...)` will be undefined resulting the `.close()` is not a function error. So, modifying the code to check if the notification is available and only call `$timeout(...)` when available as below:

```js
      if (missingRequiredFields.length) {
          var notification = CRM.alert(...);

          notification && $timeout(function () {
            notification.close();
            ...
          }, 5000);
          ...
        }
```

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
+++ b/uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
@@ -362,8 +362,8 @@ define([
           var notification = CRM.alert(missingRequiredFields.join(', '),
             missingRequiredFields.length === 1 ? 'Required field' : 'Required fields', 'error');
 
-          $timeout(function () {
-            notification && notification.close();
+          notification && $timeout(function () {
+            notification.close();
             notification = null;
           }, 5000);
 
--- a/uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
+++ b/uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
@@ -361,10 +361,12 @@ define([
         if (missingRequiredFields.length) {
           var notification = CRM.alert(missingRequiredFields.join(', '),
             missingRequiredFields.length === 1 ? 'Required field' : 'Required fields', 'error');
+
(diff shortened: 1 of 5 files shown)


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -361,10 +361,12 @@ define([
         if (missingRequiredFields.length) {
           var notification = CRM.alert(missingRequiredFields.join(', '),
             missingRequiredFields.length === 1 ? 'Required field' : 'Required fields', 'error');
+
           $timeout(function () {
-            notification.close();
+            notification && notification.close();
```
deb1990: Every change needs to be documented. I am not sure what this change is about. Add details about it in the description.

bmnepali: Updated PR description.

deb1990: The whole `$timeout` should not be called if `notification` is not present, so a if condition can be added before the $timeout

bmnepali: done

PR #224, opened 2017-06-27
PCHR-2135: Timeline stuck while assignment type was changed to another

bmnepali:
## Overview
Currently, when selected a different Task List in Assignment Modal in T&A, and on trying to change the Assignment Type, does not update the Task List selected value appropriately. 
To fix this issue, a  scope variable is create to store the selected activity's data and use it to display selected task list in "Select Task List" dropdown.  And use the selected data to filter the tasks for the selected assignment type.

## Before
The selected task list type (activity type) does not get updated on switching between assignment types.
timelinestuck

## After
The selected task list type (activity type) gets updated on switching between assignment types.
after_2135-min

## Technical Details
The selected option "Task Type List" in ui-select did not get updated when Assignment type was changed. This was a issue being the scope inheritance, due to which the selected value is not in sync.

This issue is fixed by using ` . (dot)` notation  for ng-model, so the variable that changes is not attached directly to the `scope`.

1. Wrap `$scope.activitySet` by new $scope variable `$scope.activity` as following, 
```js
$scope.activity = {
     activitySet: {}
};
```
2. When `setData()` is called setting `$scope.activity.selected` and `$scope.activitySet` respectivly with selected activity data object.
```js
$scope.setData = function () {
    ...
    $scope.activity.activitySet = assignmentType.definition.activitySets[0];
    ...
};
```
3. Use nested object `activity.selected` in `ng-model` as following in `/uk.co.compucorp.civicrm.tasksassignments/views/modal/assignment.html` file.
```html
<ui-select ... ng-model="activity.activitySet"  ...>
    ...
</ui-select>
```
4. Changed every occurrence of `$scope.activitySet` to `$acope.activity.activitySet` in both  Assignment Modal Controller and view.

### Update 1:
The issue was reopened for the flickring issue after adding the fix. And concluded with the point as the expected behavior, with few left and optimization changes.
1. Change `<hr class="separately" ng-show="activitySet.name && +settings.tabEnabled.documents" />` to `<hr class="separately" ng-show="activity.activitySet.name && +settings.tabEnabled.documents" />`

2. Use `ng-show` instead of `ng-if` in document list block in `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js`

i.e  change `<div ng-g="..." class="row">` to `<div ng-if="..." class="row">`

- [x] Tests Pass


igorpavlov:
Please improve the Overview to the PR, it should describe what has been done in the PR, but not the issue itself. The issue should be noted in the Before section along with GIF.

igorpavlov:
In Technical Details please change phrasing "can be fixed" with "is fixed".

bmnepali:
@igorpavlov Updated PR descriptions and technical details.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
```diff
@@ -49,6 +52,9 @@ define([
             $scope.activitySet = {};
             $scope.taskList = [];
             $scope.documentList = [];
+            $scope.activity = {
```
igorpavlov: I am not sure I understand why it should be nested, despite of the fact that you have mentioned it in the PR description. Could you please provide an explanation or a reference link that recommends to use nested objects for the scope.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
```diff
@@ -147,7 +153,8 @@ define([
                     return;
                 }
 
-                $scope.activitySet = assignmentType.definition.activitySets[0];
+                $scope.activity.selected = assignmentType.definition.activitySets[0];
```
igorpavlov: I am not sure I understand the code. The `$scope.activitySet` is defined as {} at the start and `$scope.activity.selected` as []. Then you assign something to `$scope.activity.selected` and then back to `$scope.activitySet`? There must be an explanation why it works and probably there could be a better potential implementation. Also I am not sure it is a good idea to ng-repeat through {} at all.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
bmnepali: The main propose of using nesting variables is to use the .(dot) notation in ng-model. Taking refrence form following  links:
- https://stackoverflow.com/questions/18128323/if-you-are-not-using-a-dot-in-your-angularjs-models-you-are-doing-it-wrong
- http://learnwebtutorials.com/why-ng-model-value-should-contain-a-dot

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
bmnepali: I have simplified the solution, refactored and updated the PR.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
```diff
@@ -335,6 +340,7 @@ define([
                 }
 
                 angular.copy(documentList, $scope.documentList);
+
```
igorpavlov: Is there a need of extra line here?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
```diff
@@ -294,9 +299,9 @@ define([
               });
             };
 
-            $scope.$watch('activitySet', function (activitySet) {
+            $scope.$watch('activity.activitySet', function (activitySet) {
```
igorpavlov: Is there a need of extra line here?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.js
bmnepali: no need! will remove it.

bmnepali:
Fixed the merge conflicts by rebase!

PR #217, opened 2017-06-15
PCHR-2321: Making document modal (due date) field non required

omarabuhussein:
## Overview
Document (due date) field is currently marked as required , but it shouldn't be. 

## Before
Creating a document require a (due date) field to be set
before

## After
Creating a document  do not require a (due date) field to be set
after

and since now the (due date) can be empty (null) for documents , I've updated the due date filters logic to handle null values so documents with null due dates will appear only if both (from) and (until) date filters are empty or if (from) is set but (until) is empty.

## Technical Details
The document API  create end point doesn't require Due Date (activity_date_time) but at the front end it is marked as required , so I had to remove required  = true as well as  required field indicator for that field .

I also had to remove the js code that checks for the field if it is set and show an error if it doesn't from the document modal controller.

Also even after marking this field as non-required it was by default set the (due date) to the date in which the document is created which is changed to ensure that it will be null.

---

- [x] Tests Pass

PR #225, opened 2017-06-29
PCHR-2321: Making document modal (due date) field non required

omarabuhussein:
Original PR : https://github.com/compucorp/civihr-tasks-assignments/pull/217

This is a mirror for the above PR but branched from staging instead of master since it is not going to be a hotfix anymore .

### Update:
The issue was reopened because, when editing existing and document with no empty due date as empty, the document does get saved with empty due date rather contains the previous due date.

This PR fixes the issue by setting the `activity_due_date` to `""` rather than `null` in `$scope.confirm()` function in file `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js` as follows:
Change from:
```js
doc.activity_date_time = $scope.parseDate(doc.activity_date_time) || null;
```
To: 
```js
doc.activity_date_time = $scope.parseDate(doc.activity_date_time) || "";
```
### Comments
- Added the test cases for the added changes.

- [x]  Test passed


uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-document_test.js
```diff
@@ -174,6 +175,7 @@ define([
 
     describe('isRole()', function () {
       beforeEach(function () {
+
```
igorpavlov: I think you need to remove that empty line and put an empty line after the initController(); because the next line is assignment.

bmnepali:
Rebased commits to fix the merge conflicts after pr review passed.

PR #230, opened 2017-07-10
PCHR-2357: Task dashboard target and assignee names show up as blank on page refresh

bmnepali:
## Overview
When there are large no of tasks and documents, In T&A tasks dashboard, the contact associated with those tasks (most of the contact ids are duplicated) are fetched using get request and for large not of contact ids (without filtering the unique ids) in URL, server rises the `414 Request URI too large` exception resulting the contacts not being fetched.

To fix this issue, before  get call, filtering the unique contact ids and then only make the get request, so the the no of contact ids to send will be less in number and eliminates the chance of getting large URL.

## Before
Shows 414 console error for request url being larger and fetching the contacts gets failed.
screen shot 2017-07-10 at 8 36 16 pm

Using GET request to send everything attached as query string causing to make the url large, which throws the 414 error.
screen shot 2017-07-10 at 8 42 26 pm

## After
Now the request URl is limited in length and only the unique contact ids are sent in the URL. So, this should never trigger 414 error, resulting success fetching of requested contacts.
screen shot 2017-07-11 at 3 33 35 pm

## Technical Details
1. Created a separate service called `Contact`, in  file `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/contact.js`.
```js
services.factory('Contact', ['$resource', '$httpParamSerializer', 'config', '$log', function($resource, $httpParamSerializer, config, $log) {
    $log.debug('Service: Contact');

    return $resource(config.url.REST, {
      action: 'get',
      entity: 'contact',
      debug: config.DEBUG
    });
  }]);
```
Here, the `$resource` is separated and kept as a  separate  factory.

2. Calling the `Contact.get()` in `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/contact.js` when 'ContactService.get()` is called.
```js
services.factory('ContactService', ['Contact', '$q', '$filter', '$rootScope', 'UtilsService', '$log',
    function (Contact, $q, $filter, $rootScope, UtilsService, $log) {
      ...
      return {
        get: function (id) {
         ...
         // Get only the ubique contact ids to be fetched.
         id = {IN: _.uniq(id.IN)};

         Contact.get({ action: 'get'}, {
            json: {
              id: id,
              return: 'display_name, sort_name, id, contact_id, contact_type, email'
            }
         }, function (data) {
      ...
   })
```
5. Added test cases for the new changes in ` `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/contact.js` service in file `uk.co.compucorp.civicrm.tasksassignments/js/test/services/contact_test.js`.

## Comments
- Modified the related test cases which required to update after the change in `uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/contact.js`
- Organized the Mock data for contact

---

- [x] Tests Pass


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/contact.js
```diff
@@ -26,11 +31,13 @@ define([
 
           var deferred = $q.defer();
 
+          // Get only the ubique contact ids to be fetched.
```
deb1990: typo in `ubique `

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/contact.js
```diff
@@ -26,11 +31,13 @@ define([
...
+          id = {IN: _.uniq(id.IN)};
```
deb1990: Fix the spacing please `id = { IN: _.uniq(id.IN) };`

uk.co.compucorp.civicrm.tasksassignments/js/test/services/contact_test.js
```diff
@@ -0,0 +1,61 @@
...
+      spyOn(Contact, 'get').and.callThrough();
+
+      promise = ContactService.get(jsonData.json.id);
+    });
+
+    afterEach(function () {
+      $httpBackend.flush();
+    });
+
+    describe('get()', function () {
```
deb1990: Could you add a test to check if unique values are getting sent

bmnepali: Added test case.

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document-list_test.js
```diff
@@ -1,3 +1,4 @@
+/* globals inject */
```
igorpavlov: Are you sure this is needed?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/contact.js
```diff
@@ -1,118 +1,127 @@
...
+
+      return {
+        get: function (id) {
+          if (!id || (typeof +id !== 'number' && typeof id !== 'object')) {
+            return null;
+          }
+
+          var deferred = $q.defer();
+
+          // Get only the ubique contact ids to be fetched.
```
igorpavlov: Typo, should be "unique"

uk.co.compucorp.civicrm.tasksassignments/js/test/services/document_test.js
```diff
@@ -81,12 +81,12 @@ define([
...
 
     beforeEach(function () {
-      spyOn(ContactService, 'get').and.callFake(fakePromise(contactMock.contactList));
+      spyOn(ContactService, 'get').and.callFake(fakePromise(contactMock.list()));
       spyOn(AssignmentService, 'get').and.callFake(fakePromise(assignmentMock.assignmentList));
       spyOn(ContactService, 'updateCache').and.returnValue(deferred.promise);
       spyOn(AssignmentService, 'updateCache').and.returnValue(deferred.promise);
 
-      $httpBackend.whenGET(/action=get&entity=contact/).respond({});
+      $httpBackend.whenGET(/action=get&debug=true&entity=contact/).respond({});
```
igorpavlov: Just wondering, could it be temporary `debug=true` (fine if not)?

bmnepali: It can not be temporary, it is always attached to url indicating whether to get the data in debug mode or  not.

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document-list_test.js
bmnepali: yeah! its needed or pre-commit hook catches it.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/contact.js
bmnepali: fixed!

uk.co.compucorp.civicrm.tasksassignments/js/test/services/document_test.js
igorpavlov: Could you explain please why it needs to contain now debug=true?

igorpavlov: Please also provide any other existing examples of the similar code if any.

igorpavlov: > Currently, its set to true because in file /uk.co.compucorp.civicrm.tasksassignments/js/test/globals.js the debug: 1 is set so when running the tests it will always be true, so it wont get interfere with the local environment setup.

Got it now. Thanks.

bmnepali: I updated the comment above.

igorpavlov: Is there a specific need to have debug set to 1 in globals?

bmnepali:
@igorpavlov  I have created a rough report here:
There are two different implementation of $resource in making http calls  in T&A 
A. Implementation One: As in file uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
```js
return $resource(config.url.REST, {
  'action': 'get',
  'entity': 'Task',
  sequential: 1,
 debug: config.DEBUG
}
```
So, when we make http call, debug is attached directly in url as query string like in “/civicrm/ajax/rest?action=get&debug=false&entity=Task”. The debug can’t be avoided and its value can either be true or false.

B. In another way, debug is passed in “json” object like in file uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/settings.js
```js
Settings.get({
  json: {
   'sequential': '1',
   'debug': config.DEBUG,
   'fields': fields
  }
}, function(data){
   …
});
```
Where the debug value is passed in `json`, which is then attached in the url as query string and made a http request, there is no debug directly attached in the url. i.e. the url now becomes  similar to “civicrm/ajax/rest?action=get&entity=Assignment” in which case “debug=true” can be avoided.

I have checked the test files and the implementation of $resource in services T&A, since these both implementation exists, its not possible to remove “debug=ture/false” in test cases. 

The main reason in setting `debug:1` in “uk.co.compucorp.civicrm.tasksassignments/js/test/globals.js” is to enforce the standard value for debug (which is always true) while we run tests and it does not depend on the debug status in our local dev environment.

uk.co.compucorp.civicrm.tasksassignments/js/test/services/contact_test.js
```diff
@@ -0,0 +1,61 @@
...
+    $httpBackend.whenGET(/action=get&entity=CaseType/).respond({});
+    $httpBackend.whenGET(/action=get&debug=true&entity=Task/).respond({});
+  };
+
+  describe('ContactService', function () {
+    var Contact, ContactService, $httpBackend, promise, jsonData;
+
+    beforeEach(module('civitasks.appDashboard'));
+
+    beforeEach(inject(function (_Contact_, _ContactService_, _$httpBackend_, _$q_) {
```
igorpavlov: Not sure _$q_ is needed here.

uk.co.compucorp.civicrm.tasksassignments/js/test/services/document_test.js
bmnepali: Nah! it simply setting the value for debug when we run tests.

uk.co.compucorp.civicrm.tasksassignments/js/test/services/contact_test.js
bmnepali: fixed

PR #234, opened 2017-07-18
PCHR-2309: Target contact and notification shown wrong in SSP

bmnepali:
## Overview
The document modal in SSP is being reused form T&A. Currently, the target contact is displayed in disabled dropdown in Document modal, when viewed form SSP by manager. The target contact should be displayed in a plain text as displayed in the wireframe.

## Before and After
The respective before and after screenshots can be found in following PRs:
1. https://github.com/compucorp/civihr-employee-portal/pull/327
2. https://github.com/compucorp/civihr-employee-portal-theme/pull/192

## Technical Details
1. Added a <p> where the target contact is displayed in plain text, which can be viewed form SSP.
```html
 <p class="target-contact" ng-if="isRole('manager')">{{contacts.target[0].label}}</p>
```
Here,  `contacts.target[0].label` is good to use as `contacts.target` hold a single target contact when opened a document in edit mode.

2. To display the notification in SSP when there are validation errors in Document on save or update. A `notification` service is used to display the error.
```js
if (missingRequiredFields.length) {
          var notificationTitle = missingRequiredFields.length === 1 ? 'Required field' : 'Required fields';
          var missingFields = missingRequiredFields.join(', ');
          var alert = notification.alert(notificationTitle, missingFields);

          alert && $timeout(function () {
            alert.close();
            alert = null;
          }, 5000);

          return false;
        }
```

- [x] Tests Pass


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -379,13 +379,24 @@ define([
         }
 
         if (missingRequiredFields.length) {
-          var notification = CRM.alert(missingRequiredFields.join(', '),
-            missingRequiredFields.length === 1 ? 'Required field' : 'Required fields', 'error');
+          var notification;
```
AkA84: put the entire notification logic in a function

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -379,13 +379,24 @@ define([
...
+          var missingFields = missingRequiredFields.join(', ');
+
+          if ($scope.isRole('manager')) {
+            $rootScope.$broadcast('ssp:notify', {
+              type: 'error',
+              title: notificationTitle,
+              body: missingFields
+            });
+          } else {
+            notification = CRM.alert(missingFields, notificationTitle);
```
AkA84: the `CRM.alert` doesn't exist in the ssp no? (i imagine) So instead of checking by role, check if `CRM.alert` exists or not, and if it doesn't use toastr

AkA84: also, the notification (whether via toastr or `CRM.alert`) should happen only in one place. If you are sending an event (ssp:notify), then the notification logic should be only in the place that catches the event. So i think the logic should be changed. There should be only one place that triggers the alert

bmnepali: `CRM.alert()` is available in SSP. The problem is that the way it works in SSP.  Here
```js
/**
   * @see https://wiki.civicrm.org/confluence/display/CRMDOC/Notification+Reference
   */
  CRM.alert = function (text, title, type, options) {
    debugger
    type = type || 'alert';
    title = title || '';
    options = options || {};
    if ($('#crm-notification-container').length) {
      debugger
      var params = {
        text: text,
        title: title,
        type: type
      };
      // By default, don't expire errors and messages containing links
      var extra = {
        expires: (type == 'error' || text.indexOf('<a ') > -1) ? 0 : (text ? 10000 : 5000),
        unique: true
      };
      options = $.extend(extra, options);
      options.expires = options.expires === false ? 0 : parseInt(options.expires, 10);
      if (options.unique && options.unique !== '0') {
        $('#crm-notification-container .ui-notify-message').each(function () {
          if (title === $('h1', this).html() && text === $('.notify-content', this).html()) {
            $('.icon.ui-notify-close', this).click();
          }
        });
      }
      debugger
      return $('#crm-notification-container').notify('create', params, options);
    }
    else {
      if (title.length) {
        text = title + "\n" + text;
      }
      alert(text);
      return null;
    }
  };
```
The `CRM.alert()` looks for the element with id `#crm-notification-container` which is not present in SSP so it simply displays the default alert.

deb1990: @bmnepali As in SSP we dont have `#crm-notification-container`, it might be easier to add that markup in ssp and style it, so that we dont need to use toaster.
cc @AkA84

bmnepali: @deb1990 I've already proposed with the approach you mentioned in technical technical description of the ticket. And it needs few changes in backend code. Having discussion with @AkA84, we decided to find alternative solution and came to use toastr.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.js
```diff
@@ -379,12 +380,13 @@ define([
...
-            missingRequiredFields.length === 1 ? 'Required field' : 'Required fields', 'error');
+          var notificationTitle = missingRequiredFields.length === 1 ? 'Required field' : 'Required fields';
+          var missingFields = missingRequiredFields.join(', ');
+          var alert = notification.alert(notificationTitle, missingFields);
 
-          notification && $timeout(function () {
-            notification.close();
-            notification = null;
+          alert && $timeout(function () {
+            alert.close();
```
deb1990: Why do we need to do this? `CRM.alert`supports a parameter for the fadeout time

> expires: how long to display this message before fadeout (in ms)
set to 0 for no expiration
defaults to 10 seconds for most messages, 5 if it has a title but no body, or 0 for errors or messages containing links

bmnepali: Thanks for catching! I think should pass the {expires: 5} options instead to `notification`.

PR #237, opened 2017-07-26
PCHR-2277: Don't Create Extra Revisions on Document Updates

MiyaNoctem:
## Overview
Updating a document sometimes created one or two revisions. It should behave the same way as activities, only creating revisions if it is set to a case.

## Before
The implementation of create API call to Documents entity was generating up to two revisions per call, when the following conditions were met:

- Params array contained 'case_id'
- Params array contained a custom field

This happened because, first, Document create API generated a new revision from the document's old data, if a value for 'case_id' was found in params array, making any updates on the new revision after. Once the initial updates were done, it checked if there were any custom fields being updated, and if
so, made an API call to Activity's create action to update the document's custom fields. This call also checks for case_id and creates a new revision if it is set in params array, or if the relationship is found in the database for the activity.

## After
The second API call to Activity's create action was probably added because custom fields were not being updated by the call to Document BAO's create method. Now, the reason for this was in order to update custom fields, API expected the physical column names to be used, instead of their field names.

Fixed by replacing custom field names for the column names and looking for custom fields defined for both Documents and Activities, before starting all updates. Also removed the second API call to Activity create action, as it is no longer needed.

With this, API call to Document's create action behaves exactly as Activity's create call, creating only one revision per update for documents assigned to a case.

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
@@ -34,8 +34,17 @@ function civicrm_api3_document_create($params) {
     return $errors;
   }
 
+  // Processing for custom data
+  $customFields = _civicrm_api3_document_getcustomfieldsflipped();
+  foreach ($customFields as $key => $value) {
+    if (isset($params[$key])) {
+      $params[$value] = $params[$key];
+    }
+  }
+
   $values = $activityArray = array();
   _civicrm_api3_custom_format_params($params, $values, 'Document');
+  _civicrm_api3_custom_format_params($params, $values, 'Activity');
 
   if (!empty($values['custom'])) {
     $params['custom'] = $values['custom'];
@@ -46,6 +55,11 @@ function civicrm_api3_document_create($params) {
   $case_id           = '';
   $createRevision    = FALSE;
   $oldActivityValues = array();
+
+  // Lookup case id if not supplied
+  if (!isset($params['case_id']) && !empty($params['id'])) {
+    $params['case_id'] = CRM_Core_DAO::singleValueQuery("SELECT case_id FROM civicrm_case_activity WHERE activity_id = " . (int) $params['id']);
+  }
   if (!empty($params['case_id'])) {
     $case_id = $params['case_id'];
     if (!empty($params['id'])) {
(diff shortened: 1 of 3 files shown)


uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
```diff
@@ -34,8 +34,17 @@ function civicrm_api3_document_create($params) {
     return $errors;
   }
 
+  // Processing for custom data
+  $customFields = _civicrm_api3_document_getcustomfieldsflipped();
+  foreach ($customFields as $key => $value) {
+    if (isset($params[$key])) {
+      $params[$value] = $params[$key];
```
omarabuhussein: are you sure that $params[$value]  is correct assignment, because if I am following the code correctly $customFields array should be in this format : 

array (
'custom_field_name_1' => 0,
'custom_field_name_2' => 1,
'custom_field_name_3' => 2,
..etc
)

so $params[0] = $params['custom_field_name_1'] doesn't seems like something make sense since $params[0] represents nothing here.

omarabuhussein:
also since the the API function contain the business logic here , and I am not sure if it possible  here but  if yes , can you add some tests to this api function ?

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
MiyaNoctem: $customFields is an array of the form:
```
array(
  'custom_field_name' => 'custom_field_column',
);
```
This way, for example, if $params['remind_me'] is set (which is one of the custom fields used for documents), it will add something like $params['custom_4'] = $params['remind_me']. Now that $params['custom_4'] is set, when _civicrm_api3_custom_format_params() runs it will identify the field as a custom field and create an entry in $values array, which will then be added to $params['custom'], which is what will be finally be used in Activity's BAO to update custom fields.

https://github.com/civicrm/civicrm-core/blob/master/CRM/Activity/BAO/Activity.php#L521

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
```diff
@@ -304,6 +304,46 @@ public function testDocumentCloningOnlyClonesCurrentRevision() {
...
+        'expire_date' => date('Y-m-d'),
+        'remind_me' => 1,
+        'status_id' => CRM_Tasksassignments_BAO_Document::STATUS_APPROVED,
+      ]);
+    }
+
+    $activityCount = civicrm_api3('Activity', 'getcount');
+    $this->assertEquals(3, $activityCount);
+
+    $this->assertDocumentRevisionCount($documents['standAloneDocument']['id'], 1);
```
omarabuhussein: I think it is better to change assertDocumentRevisionCount to getDocumentRevisionCount and do the assertion inside testOnlyDocumentsAssignedToACaseCreateRevisionsOnUpdate itself

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
```diff
@@ -304,6 +304,46 @@ public function testDocumentCloningOnlyClonesCurrentRevision() {
...
+    $this->assertDocumentRevisionCount($documents['caseDocument']['id'], 2);
+  }
+
+  private function assertDocumentRevisionCount($documentID, $expectedCount) {
+    $result = civicrm_api3('Activity', 'get', [
+      'sequential' => 1,
+      'activity_type_id' => "VISA",
```
omarabuhussein: using "VISA" here is a bit vague, It's not clear where it came from. also, is it really needed?

PR #238, opened 2017-07-28
PCHR-2363: Fix Task Reminder Sending on Contact's Task Tab

MiyaNoctem:
## Overview
Admin was no longer able to manually trigger reminder emails from the Actions ellipse on Tasks (collapsed view). Clicking on the Send Reminder option showed a modal form to add a message. When the modal form was submitted, an error was thrown, saying "Please specify 'activity_id' value.".

## Before
Processing of sendreminder modal form was not sending the activity ID as a parameter, which caused an error to be thrown by the back-end, as this is a required value.

before

## After
Fixed by adding the parameter to the corresponding service call and updating
minified js file.

after

---

- [x] Tests Pass


igorpavlov:
@MiyaNoctem could you please add GIFs to the Before and After sections? Please include inspector window in GIFs showing the error being thrown in Before and not thrown in After.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/services/task.js
```diff
@@ -231,6 +231,7 @@ define([
 
           Task.save({ action: 'sendreminder' }, {
             json: {
+              activity_id: taskId,
               notes: notes
             } || {}
```
igorpavlov: I am not quite sure why we need that ` || {}`. Could you explain please?

igorpavlov: @MiyaNoctem I don't think `|| {}` is needed here, please remove it.

MiyaNoctem: I checked with @bmnepali, and he agrees there is probably no need to keep this.

PR #248, opened 2017-08-21
PCHR-2520: Allow Inline Download of Files

mickadoo:
## Overview
We would like to allow users to view files in the browser. This functionality is not supported by the CiviCRM file endpoint, but with [this PR](https://github.com/civicrm/civicrm-core/pull/10883) introduces that functionality. To take advantage of it the links in the file modal can be changed to point to this endpoint with the new `download` parameter appended.

## Before
Clicking on files in the documents modal would display a "popup blocked warning" from Chrome.

## After
Supported files are opened in a new tab. 

## Comments
This PR is pending the approval and merge of [CRM-21086](https://github.com/civicrm/civicrm-core/pull/10883).

---

- [ ] Tests Pass
Running tests causes an error on setup about "ta_settings" being an invalid `option_group_id`


mickadoo:
This branch was so out of date since the 1.7 merge it was easier to recreate the changes in a new branch. Closing in favor of #280

PR #280, opened 2017-11-09
PCHR-2520: Allow Inline Document File Downloads

mickadoo:
## Overview
We would like to allow users to view files in the browser. This functionality is not supported by the CiviCRM file endpoint, but with [this PR](https://github.com/civicrm/civicrm-core/pull/10883) introduces that functionality. To take advantage of it the links in the file modal can be changed to point to this endpoint with the new `download` parameter appended.

## Before
Clicking on files in the documents modal would display a "popup blocked warning" from Chrome.

pchr-2520-before

## After
Supported files are opened in a new tab. 

pchr-2520-after

---

- [x] Tests Pass

Both PHPUnit and JS tests all pass (Thanks @deb1990 for helping set up headless chrome test)


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.controller.js
```diff
@@ -69,6 +69,7 @@ define([
     vm.showStatusField = showStatusField;
     vm.searchContactAssignments = searchContactAssignments;
     vm.viewFile = viewFile;
+    vm.addQueryParam = addQueryParam;
```
deb1990: Please define function alphabetically, so it should be on line 51.

deb1990:
@mickadoo Please attach some before and after screenshots.
Also , should not you fix `Running tests causes an error on setup about "ta_settings" being an invalid option_group_id`?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-document.controller.js
```diff
@@ -585,5 +586,20 @@ define([
...
+    /**
+     * Appends a query parameter to a path. Does not check if parameter is
+     * already set.
+     *
+     * @param {string} path
+     * @param {string} name
+     * @param {string} value
+     * @returns {string}
+     */
+    function addQueryParam(path, name, value) {
```
deb1990: Should be defined alphabetically. That is, line no 99

PR #282, opened 2017-11-20
PCHR-2886: Applied new colours to task and assignment screens

mukeshcompucorp:
## Overview
This PR changes colors of extension based on WCAG AA color palette of shoreditch theme.
https://invis.io/HWECY91E2#/261993810_Colour

### Tasks Before
image

### Tasks After
image

### Documents Before
image

### Documents After
image

### Calendar Before
image

### Calendar After
image

### Key Dates Before
image

### Key Dates After
image

### Before Add New Task
image

### After Add New Task
image

### Before Add New Document
image

### After Add New Document
image

### Fixed buttons
This PR also fixes wrong button glitches such as default buttons styled to look like outline buttons,
**Default button which looks like outlined button:**
image
**Outlined button**
image

Visually it does makes minimal difference though, logically class button markup should be
`<button type="button" class="btn btn-secondary-outline">Cancel</button>`
instead of `<button type="button" class="btn btn-default">Cancel</button>`

### Fix document modal property
changed variable from `showFieldAssignment` to `showAssignmentField` to increase readability and match backstopjs script like below:
```
showField: function (fieldName) {
  var casper = this.casper;

  casper.then(function () {
    casper.click(this.modalRoot + ' a[ng-click*="show' + fieldName + 'Field"]');
  }.bind(this));

  return this;
},
```
at https://github.com/civicrm/civihr/blob/staging/uk.co.compucorp.civicrm.hrcore/backstop_data/page-objects/modals/document.js#L28

Document modal field property follows pattern "show" + fieldName + "Field". so, we are making it proper to have like other fields (i.e. `showAssigneeField`)
Here the idea is to have same pattern for field is to allow function of backstopjs like `showField('Assignee')` (above function) to pass just `fieldName` and work in same logic.

It is been tried to fix this problem by changing BackstopJS script but, that requires to change other fields like, `showAssigneeField` as given in https://github.com/compucorp/civihr-tasks-assignments/blob/staging/uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html#L232 so, this changes is done as minimum change required to handle all logic without breaking anything else in extension.

---

- [x] Tests Pass


AkA84:
@mukeshcompucorp 

i'm not sure i understand that bit about the button default "looking like" the outlined button. A default button was simply (as far as i remember) a button with white background, it wasn't supposed to look like outlined buttons at all. Are you sure about this change?

also what's with [this commit](https://github.com/compucorp/civihr-tasks-assignments/pull/282/commits/cf63b6ce8881b6267e43a300dda695e5f3848d1e)? i don't see any mention in the PR

mukeshcompucorp:
@AkA84 , 
For buttons, yes, in every place we use secondary outline button.
Also, the style of default button should be white background without borders (as we have in style guide).
Current design of the button was wrong as well as it should be outlined button as per other pages we are styling throughout CiviHR or CiviCRM

For https://github.com/compucorp/civihr-tasks-assignments/pull/282/commits/cf63b6ce8881b6267e43a300dda695e5f3848d1e I've updated description now.

mi-kon:
Please hold off merging fix for datepickers @mukeshcompucorp @AkA84 !

mi-kon:
@mukeshcompucorp @AkA84 please hold off merging this width fix.

PR #287, opened 2017-11-30
PCHR-2572: Fix Task and Assignment dashboard key dates filter issue

bmnepali:
## Overview
This PR fixes the issue with while filtering key dates by date type in Task and Assignment Dashboard. Now only key dates of selected types will be displayed.

## Before
- Filtering the key dates does not display only key dates of selected type.
2572-before-min

## After
- Only key dates of selected types will be displayed.
2572-after-min

- Fixed duplicate data being appearing on filtering date-range  filter:
2572-finnal-min

## Technical Details
1. Currently the filter filters the array in `dateContactList` property of input array.
```js
[
  {
     'date': '2017-01-07',
     'dateContactList': [
        {
           'contact_id': '93',
           'contact_external_identifier': '',
           'contact_name': 'Díaz, Kacey',
           'keydate': '2017-01-07',
           'type': 'birth_date'
        }
      ]
   }
]
```
2. Update `filterByKeyDateType` to filter the key dates even if the input key date array does not contain the `dateContactList` property holding the key dates.

So, the filter is updated to filter the data in following form as well.
```js
[
   {
      'contact_id': '93',
      'contact_external_identifier': '',
      'contact_name': 'Díaz, Kacey',
      'keydate': '2017-01-07',
      'type': 'birth_date'
    },
    ....
]
```

Here is the updated filter code:

```js
/**
 * Filters the array of key dates in both cases
 * case 1: when the input array contains key dates in dateContactList propery
 * case 2: when the input array is itself is a list of key dates list.
 *
 * @param  {Array} dateContactList List of contact key dates
 * @param  {Array} dateTypeList list of date types
 * @return {Array}
 */
return function (dateContactList, dateTypeList) {
  var filteredDateContactList = [];

  if (!dateContactList.length || !dateTypeList.length) {
    return dateContactList;
  }

  _.forEach(dateContactList, function (contactKeyDate) {
    var nestedKeyDates = contactKeyDate.dateContactList;

    if (nestedKeyDates) {
      _.forEach(nestedKeyDates, function (singleKeyDate) {
        if (dateTypeList.indexOf(singleKeyDate.type) !== -1) {
          return filteredDateContactList.push(contactKeyDate);
        }
      });
    } else {
      if (dateTypeList.indexOf(contactKeyDate.type) !== -1) {
        return filteredDateContactList.push(contactKeyDate);
      }
    }
  });

  return filteredDateContactList;
};
```
2. Then use the filterByDateType in templates for each key date as well. As shown below in file views/dashboard/key-dates.html.
```html
<script type="text/ng-template" id="key-date.html">
  ...
  <ul class="list-group">
    <li class="list-group-item" ng-repeat="dateContact in dateObj.dateContactList | filterByDateType:filterParams.dateType"> ... </li>
  </ul>
</script>
```
After reopened:
3. When filtering by date ranges, duplicate data gets collected in the list is repeated and and angular is throwing the "duplicate values" error . It was a regression issue after i made changes on the date-type filter.

So, modify `date-type.filter.js` to avoid adding duplicate data in the list while filtering.
```js
...
if (nestedKeyDates) {
   _.forEach(nestedKeyDates, function (singleKeyDate) {
      if (dateTypeList.indexOf(singleKeyDate.type) !== -1) {
        filteredDateContactList.push(contactKeyDate);
        
        return false; // break loop important
      }
    });
}
...
```
Note: There is no need of extra unique filter to get the unique values. So the unique-filter.js ans its related changes are removed.

## Comments
-  Refactored the `filterByDateType` filter using lodash.
- Applied style guide and fixed semistandard issues.
- Added test cases

## Tests
✅ Jasmine Tests - added/passed
✅ JS distributive files - created
⏹ CSS distributive files - not needed to run
⏹ Backstop Tests - n/a
⏹ PHP Unit Tests - not needed to run


uk.co.compucorp.civicrm.tasksassignments/views/dashboard/key-dates.html
```diff
@@ -84,6 +84,7 @@ <h3 class="panel-title pull-left" ng-bind="(filterParams.date && filterParams.da
           </div>
         </div>
       </div>
+      <br/>
```
igorpavlov: Why do we have a `<br/>` here?

igorpavlov:
@bmnepali is there a mix of automatic indentation fixing + code fixes in these commits?

https://github.com/compucorp/civihr-tasks-assignments/pull/287/commits/d90a918c55b187f1a5caab603da527c9e25d5e81

https://github.com/compucorp/civihr-tasks-assignments/pull/287/commits/d6ca7f22e6a0731445083a56338af2cb10c4fd60

If yes, please specify where have you performed manual changes.

uk.co.compucorp.civicrm.tasksassignments/js/test/filters/date-type.spec.js
```diff
@@ -0,0 +1,37 @@
...
+    beforeEach(inject(function ($filter) {
+      withDateContactList = keyDatesMock.withDateContactList();
+      withoutDateContactList = keyDatesMock.withoutDateContactList();
+
+      filter = $filter('filterByDateType');
+    }));
+
+    describe('when filtering array of key dates with dateContactList property', function () {
+      it('returns key dates that are of expected date types', function () {
+        expect(filter(withDateContactList, 'birth_date').length).toEqual(2);
```
igorpavlov: Is it possible to compare with mock properties instead of fixed constants?

uk.co.compucorp.civicrm.tasksassignments/views/dashboard/key-dates.html
bmnepali: My bad! thanks for catching, I think, I left it while debugging.

bmnepali:
@igorpavlov both the commits don't have automatically fixed semistandard code issues. The reported issues after commit are manually fixed in [this commit](https://github.com/compucorp/civihr-tasks-assignments/commit/d6ca7f22e6a0731445083a56338af2cb10c4fd60) only and updated as per style guide.

igorpavlov:
@bmnepali OK, I think it would be great to fix indentation in a separate commit next time.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/date-type.js
```diff
@@ -1,38 +1,35 @@
...
-                i = 0,
-                ii,
-                inputArrLen = inputArr.length,
-                dateTypeArrLen = dateTypeArr.length,
-                dateContactList,
-                dateContactListLen;
-
-            if (!inputArrLen || !dateTypeArrLen) {
-                return inputArr;
+        return function(dateContactList, dateTypeList) {
```
igorpavlov: I think we need a description here on the function. At the moment we need to look inside the func to understand what `dateContactList` and `dateTypeList` are and what kind of data it returns.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/date-type.js
```diff
@@ -1,38 +1,35 @@
...
-                dateContactListLen = dateContactList.length;
+            _.forEach(dateContactList, function (dateContact) {
+              var dateContactList = dateContact.dateContactList;
 
-                for (ii; ii < dateContactListLen; ii++) {
-                    if (dateTypeArr.indexOf(dateContactList[ii].type) !== -1) {
-                        filteredArr.push(inputArr[i]);
-                        break;
-                    }
+              dateContactList && _.forEach(dateContactList, function (singleDateContact) {
```
igorpavlov: I think this is an example of overusing `() && ()`. It is really confusing to read something like that. Let's stick with classic `if (...) { _.foreach... }`

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/date-type.js
```diff
@@ -1,38 +1,35 @@
...
+            if (!dateContactList.length || !dateTypeList.length) {
+                return dateContactList;
             }
 
-            for (i; i < inputArrLen; i++) {
-                ii = 0,
-                dateContactList = inputArr[i].dateContactList,
-                dateContactListLen = dateContactList.length;
+            _.forEach(dateContactList, function (dateContact) {
+              var dateContactList = dateContact.dateContactList;
```
igorpavlov: `dateContactList` is your func parameter, here you are redefining it. Also `var` will remain redefined outside the scope of the function, so it is actually dangerous. Please use different names or do not define `dateContactList` at all.

igorpavlov:
Could you please explain why did you decide to create a filter and not `_.uniq` directly in the controller?

> When filtering by date ranges, some thing the data is repeated and and angular is throwing the "dublicate values" error . Create a filter called "unique " which will filter the data to return non dublicate values.

Why is the data repeated at first, what is the reason?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/unique.filter.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-env amd */
+
+define([
+  'common/lodash',
+], function (_) {
+  'use strict';
+
+  uniqueFilter.__name = 'unique';
+  uniqueFilter.$inject = ['$filter', '$rootScope', '$log'];
```
igorpavlov: Please sort alphabetically.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/unique.filter.js
```diff
@@ -0,0 +1,31 @@
...
+
+  function uniqueFilter ($filter, $rootScope, $log) {
+    $log.debug('Filter: unique');
+
+    /**
+     * Filters the array to return unique values
```
igorpavlov: "Filters *an* array..."

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/unique.filter.js
```diff
@@ -0,0 +1,31 @@
...
+     *
+     * @param  {Array} inputArray
+     * @param  {Array} identifier to filter by
```
igorpavlov: Doesn't uniq accept only one param? https://lodash.com/docs/4.17.4#uniq

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/unique.filter.js
```diff
@@ -0,0 +1,31 @@
...
+     * @return {Array}
+     */
+    return function (inputArray, identifier) {
+      if(!inputArray.length) {
```
igorpavlov: `_.uniq([])` returns `[]`

Why this check is needed?

uk.co.compucorp.civicrm.tasksassignments/js/test/filters/unique.filter.spec.js
```diff
@@ -0,0 +1,29 @@
...
+
+      beforeEach(module('tasks-assignments.dashboard'));
+      beforeEach(inject(function ($filter) {
+        withDateContactList = keyDatesMock.withDateContactList();
+        withoutDateContactList = keyDatesMock.withoutDateContactList();
+
+        filter = $filter('unique');
+      }));
+
+      describe('when filtering array of data by a date', function () {
```
igorpavlov: This test is not enough. This filter is a pure function, a straightforward algorithm, it needs to be tested similar to this: https://github.com/civicrm/civihr/pull/2238/files#diff-4f7c3ab98c5b2837916a2da49d5b4894R30

```js
// semi-pseudo code

var testCases = [
  // [input, expectedResult, uniqueBy]
  // Common cases
  [['a', 'b', 'c'], ['a', 'b', 'c']],
  [['a', 'd', 'a', 'd', 'c', 'a'], ['a', 'd', 'c']],
  [[1, 2, 3, 3, 1, 2], [1, 2, 3]],
  [[{ id: 1 }, { id: 2 }, { id: 1 }], [{ id: 1 }, { id: 2 }], 'id'],

  // Edge cases
  [[], []],
  ['hello world', 'helo wrd'],
  [{}, ????], // not sure what should happen here tho
  [[undefined, undefined], [undefined]], // not sure tho, may be it should be just []
  [[{ id: 1 }, { id: 2 }, { id: 1 }], [{ id: 1 }, { id: 2 }, { id: 1 }], 'date'],
];

testCases.forEach(function (testCase) {
  describe('...', function () {
    it('transforms ' + JSON.stringify(testCase[0]) + ' to ' + JSON.stringify(testCase[1]), function () {
      expect(filter(testCase[0])).toEqual(testCase[1]);
    });
  });
});
```

igorpavlov: @bmnepali forgot to mention, I think it would be expected to see more test cases than presented in the snippet above. The snippet is just for the info.

bmnepali:
@igorpavlov  I created a `filter` instead of using `_.uniq()` for following reasons:
1. there might be instances where we might need to filter data based on the given property in the template itself. 
2. And by creating `filter` it  will be available across the extension whereas creating within the controller will be limited to the specific controller.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/unique.filter.js
bmnepali: right! not needed!

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/unique.filter.js
bmnepali: We are using Lodash 3.10.1 https://lodash.com/docs/3.10.1#uniq. It can accept second param as well incase of list of objects.

igorpavlov: @bmnepali Thanks, I see now. That means when we update lodash we have to be sure our tests cover this. Otherwise we will have instant regressions and they may not be easy to spot. May I ask you to check where else _.uniq is used in the extension with 2 params?

In lodash 4 _.uniq accepts 1 param, for additional identifier you use _.uniqBy.

bmnepali: @igorpavlov 
> It also can return false positives like _.uniq(_.map(allEntitlements, function (entitlement) { but anyway, that is for a single check. Hope it helps!

Here we are using `_.uniq()` only to find the unique `ids`, so i don't think, there are chance of getting false positive values. 
```js
var authorsIDs = _.uniq(_.map(allEntitlements, function (entitlement) {
    return entitlement.comment_author_id;
}));
```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/date-type.filter.js
```diff
@@ -33,18 +33,20 @@ define([
...
-              return filteredDateContactList.push(contactKeyDate);
+              filteredDateContactList.push(contactKeyDate);
+
+              return false; // break loop
             }
           });
         } else {
           if (dateTypeList.indexOf(contactKeyDate.type) !== -1) {
-            return filteredDateContactList.push(contactKeyDate);
+            filteredDateContactList.push(contactKeyDate);
```
deb1990: why `return` is removed?

uk.co.compucorp.civicrm.tasksassignments/js/test/filters/date-type.spec.js
```diff
@@ -20,33 +20,18 @@ define([
 
     describe('when filtering array of key dates with dateContactList property', function () {
       it('returns key dates that are of expected date types', function () {
-        expect(filter(withDateContactList, 'birth_date')).toEqual(filterDates(withDateContactList, 'birth_date'));
-        expect(filter(withDateContactList, 'period_end_date')).toEqual(filterDates(withDateContactList, 'period_end_date'));
-        expect(filter(withDateContactList, 'period_start_date')).toEqual(filterDates(withDateContactList, 'period_start_date'));
+        expect(filter(withDateContactList, 'birth_date').length).toEqual(10);
```
deb1990: 1. Previously the data was being compared, why are we removing that? 
2. If we really need to remove the same, the test case description also needs to be renamed.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/filters/date-type.filter.js
bmnepali: There is no need to return any value, as we are simply pushing the value to `filteredDateContactList`.

uk.co.compucorp.civicrm.tasksassignments/js/test/filters/date-type.spec.js
bmnepali: > Previously the data was being compared, why are we removing that?

The `filterByDates` was removed because it was not giving the actual values we want. If we want to update the function to work as we expect the data, we would have to implement the code form `date-type.filter.js`. So, I removed it.

PR #291, opened 2017-12-06
PCHR-3032: Implement AngularJS Style Guide and JS Semi-Standard across Tasks & Assignments app

AkA84:
## Overview
This PR makes the Tasks & Assignments angular app and its file to follow both the AngularJS Style Guide and the JS Semi-Standard Style

## Technical Details
Please refer to the technical details in https://github.com/civicrm/civihr/pull/2066 to get the idea behind this PR, as it does basically the same thing

### Resources services
There was this common pattern where a service would be bundled with one or more resources in the same file. Couple of examples:
```js
// services/assignment.js

services.factory('Relationship', function () {
  return $resource(config.url.REST, { /* ... */});
});

services.factory('Assignment', function () {
  return $resource(config.url.REST,  /* ... */});
});

services.factory('AssignmentSearch', function () {
  return $resource(config.url.ASSIGNMENTS + '/ajax/unclosed');
});

services.factory('AssignmentType', function () {
  return $resource(config.url.REST, { /* ... */});
});

services.factory('AssignmentService', [
  'Relationship', 'Assignment', 'AssignmentSearch', 'AssignmentType', 
  function (Relationship, Assignment, AssignmentSearch, AssignmentType) {
    // ...
  }
]);
```
```js
// services/contact.js

services.factory('Contact', function() {
  return $resource(config.url.REST, { /* ... */});
}]);

services.factory('ContactService', ['Contact', function (Contact) {
  // ...
}]);
```
To keep the one-thing-per-file approach, each resource had been moved to a separate `xyz.resource.js` file, and registered on a `tasks-assignments.resources` module

```js
// resources/contact.resource.js

define(function () {
  'use strict';

  Contact.__name = 'Contact';
  Contact.$inject = [
    '$resource', '$httpParamSerializer', 'config', '$log'
  ];

  function Contact ($resource, $httpParamSerializer, config, $log) {
    // ...
  }

  return Contact;
});
```

```js
// modules/tasks-assignments.resources.js

define([
  'common/angular',
  'tasks-assignments/resources/assignment-search.resource',
  'tasks-assignments/resources/assignment-type.resource',
  'tasks-assignments/resources/assignment.resource',
  'tasks-assignments/resources/contact.resource',
  'tasks-assignments/resources/relationship.resource'
], function (angular, AssignmentSearch, AssignmentType, Assignment, Contact, Relationship) {
  'use strict';

  return angular.module('tasks-assignments.resources', [])
    .factory(AssignmentSearch.__name, AssignmentSearch)
    .factory(AssignmentType.__name, AssignmentType)
    .factory(Assignment.__name, Assignment)
    .factory(Contact.__name, Contact)
    .factory(Relationship.__name, Relationship);
});
```

#### Problem
The `tasks-assignments.min.js` file, in the contact summary page, might end up being added twice: once for the Tasks tab and once for the Documents tab (in case the user clicks on both of them one after the other). This is due to the [pageRun hook](https://github.com/compucorp/civihr-tasks-assignments/blob/staging/uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php#L197) used in the TA extension
```php
function tasksassignments_civicrm_pageRun($page) {
  if ($page instanceof CRM_Tasksassignments_Page_Tasks ||
      $page instanceof CRM_Tasksassignments_Page_Documents) {
    CRM_Core_Resources::singleton()
      ->addScriptFile('uk.co.compucorp.civicrm.tasksassignments', CRM_Core_Config::singleton()->debug ? 'js/src/tasks-assignments.js' : 'js/dist/tasks-assignments.min.js',1010);
  }
}
```
This causes the apps of the tasks and documents tab to be bootstrap multiple times which, with the changes introduced in this PR, triggers an error
```js
Uncaught Error: [ng:btstrpd] http://errors.angularjs.org/1.5.9/ng/btstrpd?p0=%26lt%3Bdiv%20id%3D%22cividocuments%22%20ng-controller%3D%22MainController%22%20ct-spinner%3D%22%22%20ct-spinner-show%3D%22%22%20class%3D%22ng-scope%22%20style%3D%22position%3A%20relative%3B%22%26gt%3B
```

#### Before
Not 100% sure, but before this PR this didn't cause any problem because the `angular.bootstrap` method was invoked in the `'tasks-assignments/app'` module when receiving the `taInit` event from the tab's html page
```js
// tasks-assignments/app.js
document.addEventListener('taInit', function (e) {
  angular.bootstrap(document.getElementById(e.detail.module), ['civitasks.' + e.detail.app]);
});
```
and the module in turn was required by the main file, `tasks-assignments.js`:
```js
// tasks-assignment.js
(function () {
  require(['tasks-assignments/app']);
}())
```
My guess was that the  `require()` was invoked only the first time (after that, the module was already loaded), thus avoiding multiple bootstraps

#### After
In this PR though, the `tasks-assignments/app` module was deleted, and the `angular.bootstrap` method is invoked directly in `tasks-assignments.js` file. There is no `require()` anymore to shield from multiple bootstraps, which then occur

A simple solution is to check then if the module is already bootstrapped before actually proceeding to initialize it

```js
// tasks-assignments.js
document.addEventListener('taInit', function (e) {
  var element = angular.element(document.getElementById(e.detail.module));
  var isBootstrapped = element.injector();

  if (!isBootstrapped) {
    angular.bootstrap(element, ['tasks-assignments.' + e.detail.app]);
  }
});
```
Of course an even better solution would be to avoid the file to be added twice in the first place, but after a quick investigation it doesn't seem that the solution would be straightforward. Given the massive scope of this PR already (and how time consuming had already been for QA to test it), i preferred to go with a simpler solution

----

* [x] Tests pass


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments.js
```diff
@@ -23,7 +23,12 @@
     'use strict';
 
     document.addEventListener('taInit', function (e) {
-      angular.bootstrap(document.getElementById(e.detail.module), ['tasks-assignments.' + e.detail.app]);
+      var element = angular.element(document.getElementById(e.detail.module));
```
igorpavlov: Please find a better name for `element` since it is too generic.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments.js
```diff
@@ -23,7 +23,12 @@
...
+      var isBootstrapped = element.injector();
```
igorpavlov: var isBootstrapped = !!element.injector();

PR #292, opened 2017-12-13
PCHR-3027: Improvements to Tests

mickadoo:
## Overview

UnitTests were failing because of warnings, and some required fabricators were missing.

## Before

- Unit tests from CiviHR would fail because of warnings when doing getcount on an activity based entity (e.g. Document, Task)
- There were no fabricators for Task and TaskType
- Tests were failing

## After

- Unit tests from CiviHR pass
- There are new fabricators for Task and TaskType
- T&A has its own base headless test (copied from CiviHR)
- Tests pass

---

- [x] Tests Pass


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Assignment.php
```diff
@@ -41,11 +43,14 @@ private function setDefaultValues() {
    */
   private static function getOneCaseType() {
     $result = civicrm_api3('CaseType', 'get', [
-      'sequential' => 1,
       'is_active' => 1,
       'options' => ['limit' => 1],
     ]);
 
+    if ($result['count'] == 0) {
+      return CRM_HRCore_Test_Fabricator_CaseType::fabricate()['id'];
```
davialexandre: This created a dependency that we don't want to have. T&A should not depend on anything from CiviHR. It's fine to use the API here

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Task.php
```diff
@@ -0,0 +1,29 @@
...
+   * @param array $params
+   *   List of parameters to use to create the Assignment
+   *
+   * @return array
+   */
+  public static function fabricate($params = []) {
+    $params = array_merge($params, self::$defaultParams);
+    $result = civicrm_api3('Task', 'create', $params);
+
+    return array_shift($result['values']);
```
davialexandre: Given that the other fabricators in this extension (`Document` and `Assignment`) use the the BAO instead of the API, I think we should do the same here

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Task.php
```diff
@@ -0,0 +1,29 @@
...
+class CRM_Tasksassignments_Test_Fabricator_Task {
+  private static $defaultParams = [
+    'activity_name' => 'Test Task',
+    'source_contact_id' => 'user_contact_id',
+    'target_id' => 'user_contact_id',
+    'assignee_id' => 'user_contact_id',
+  ];
+
+  /**
+   * Fabricates an assignment merging given parameters with default minimum
```
davialexandre: Please update this description. I think you copied it from somewhere else.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/TaskType.php
```diff
@@ -0,0 +1,26 @@
...
+class CRM_Tasksassignments_Test_Fabricator_TaskType {
+
+  private static $defaultParams = [
+    'option_group_id' => 'activity_type',
+    'name' => 'Test Task',
+    'component_id' => 'CiviTask',
+  ];
+
+  /**
+   * Fabricates an assignment merging given parameters with default minimum
```
davialexandre: Please update this description. I think you copied it from somewhere else.

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
```diff
@@ -592,6 +592,11 @@ function civicrm_api3_document_get($params) {
             )
         ));
 
+      // foreach loop will break if doing 'is_count' request
```
davialexandre: Why?

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
```diff
@@ -15,19 +17,15 @@ class api_v3_DocumentTest extends CiviUnitTestCase {
...
-    $upgrader = CRM_Tasksassignments_Upgrader::instance();
-    $upgrader->install();
-
-    $hrCaseUpgrader = CRM_HRCase_Upgrader::instance();
-    $hrCaseUpgrader->install();
+  public function setUpHeadless() {
+    return \Civi\Test::headless()
+      ->installMe(__DIR__)
+      ->apply();
+  }
```
davialexandre: The `BaseHeadlessTest` class already implements this, so you don't need it here

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
```diff
@@ -305,12 +303,13 @@ public function testDocumentCloningOnlyClonesCurrentRevision() {
   }
 
   public function testOnlyDocumentsAssignedToACaseCreateRevisionsOnUpdate() {
+    $originalCount = civicrm_api3('Activity', 'getcount');
```
davialexandre: Instead of this, it's better to cleanup things in the `setUp` method, as it was done before

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/bootstrap.php
```diff
@@ -9,8 +9,17 @@
...
+
 require_once 'TasksassignmentsTestTrait.php';
 
+function addCiviHRExtensionToIncludePath($extensionDirectory) {
+  $civicrmRoot = realpath($GLOBALS['civicrm_root']);
+  $civiHRDir = $civicrmRoot . '/tools/extensions/civihr';
+  $extensionDir = $civiHRDir . $extensionDirectory;
+  set_include_path($extensionDir . PATH_SEPARATOR . get_include_path());
+}
+
```
davialexandre: By removing the usage of the HRCore fabricator, this won't be necessary anymore. As a rule of thumb, we shouldn't include things like this. If you need a class from another extension, please install it using the `setupHeadless` method

davialexandre:
@mickadoo also, this PR has almost nothing to do with "Gathering statistics". Please update it with a proper name that describes what it is about

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Assignment.php
mickadoo: I'll change it, but you should be aware there are other dependencies on CiviHR in T&A

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Task.php
mickadoo: It is doing something different alright. Now we have fabricators in CiviHR, some of which return and array, some of which return a BAO. In T&A they return a BAO. Before I change this can we agree on an interface for fabricators so we don't have to check them each time. I know you didn't like my suggestion to have a FabricatorInterface, but I can tell you after working with them for the last few days it would have helped me a lot. Not just for autocompletion (although there were times that would have been useful) but just so I can know what parameters are required and what the return type is.

uk.co.compucorp.civicrm.tasksassignments/api/v3/Document.php
mickadoo: Because $getResult will be an integer, I've updated the comment

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
mickadoo: I'm not sure I agree,  it could be possible that installing an extension will create an Activity. It's outside the scope of the test. I wouldn't put it in `setUp` anyway, since it is run for every test and a clean activity table is only required for this test.

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/bootstrap.php
mickadoo: I agree it's not perfect, but I don't think the solution is to install hrcore whenever we need to use test related classes from it, that just seems heavy-handed. This was a compromise, but as you mentioned we can remove it now.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Assignment.php
davialexandre: I was not aware of that. It shouldn't have happened

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Task.php
davialexandre: Let's open this discussion to other devs and see what they think, but my opinion is: For anything that is not a core entity, we should use BAOs instead of APIs, for the single reason that API calls are very slow and hurt the performance of the tests. There might be exceptions to this, as I know that some extensions have logic implemented directly at the API level (HRJobContract for example), and in that case there's no other option.

For now, let's use the BAO here

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
davialexandre: Well, the previous version of this class would delete all the activities (and a few more things) before running the tests, so my suggestion is just to keep the previous behavior.

I do think that the setUp method is the best place to add this and the reason is to remove the test everything that is not related to what is being tested. It's not always possible to do that, but here's I think it's quite simple.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Task.php
mickadoo: After discussion with Davi the decision was to continue using API calls for T&A entities as it can be difficult to find what params to pass to BAO calls for T&A.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Document.php
```diff
@@ -44,11 +43,7 @@ public static function fabricateWithAPI($params = []) {
    * @return array
    */
   private static function getDefaultParameters() {
-    $contactID = CRM_Core_Session::getLoggedInContactID();
-
-    if (!$contactID) {
-      $contactID = ContactFabricator::fabricate()['id'];
-    }
+    $contactID = self::getTestContactID();
```
davialexandre: Was this change really necessary to make the tests pass or was it an attempt to improve things and remove the usage of the Contact Fabricator from HRCore?

I think that the previous version (where a contact would be created everytime) was bad, but I feel like this one is worse, because most of the time it will probably result in an unnecessary API call to check if a contact exists in the database.

I don't think it's a good idea to set default values for fields that are FKs to other entities. For these, it's better to make it explicit which are the entities that will be linked together.

So the options here are:
- Revert the previous version, but probably replace the HRCore fabricator with the T&A that I've suggested on another comment (We can then improve this later and remove the contact fabrication altogether)
- Change the tests so that we don't set a default value for contact ID fields (source_contact_id, target_contact_id and assignee_contact_id)

Please do whichever one that is faster to implement

mickadoo: Yep, you asked me not to use CiviHR code here so I removed other references to the `ContactFabricator`. 

> I think that the previous version (where a contact would be created everytime) was bad, but I feel like this one is worse, because most of the time it will probably result in an unnecessary API call to check if a contact exists in the database.

I'd argue that a query call to check for a single contact is better than multiple queries and insertions related to the creation of a new contact.

> I don't think it's a good idea to set default values for fields that are FKs to other entities. For these, it's better to make it explicit which are the entities that will be linked together.

I'm conflicted on this myself, I really like the ability to just call `fabricate` and leave it to the defaults to decide what to use, but allowing for overriding of any of the defaults. For example for my tests for entity count in CiviHR I really don't care about any of the data, just that an entity is created.

My feeling is that creating a contact inside a document fabricator call is definitely wrong, and going down this path (of auto-creating required FK entities) will lead us into trouble later.

I can also see how using a random contact ID from the database or `user_contact_id` could lead to unexpected results, it's a matter of convenience (providing all defaults) vs clarity (requiring explicit FK IDs in `$params`) so probably better to go for the clarity option.

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
```diff
@@ -35,4 +24,26 @@ public function testDatesForDeletedContactsNotInResult() {
...
+      'last_name'    => 'Doe',
+    ];
+
+    $params = array_merge($defaultParams, $params);
+    $params['display_name'] = "{$params['first_name']} {$params['last_name']}";
+
+    $result = civicrm_api3('Contact', 'create', $params);
+
+    return array_shift($result['values']);
+  }
```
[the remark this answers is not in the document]
mickadoo: I guess so, but it just feels wrong copying and pasting code. I'll copy the fabricators from CiviHR anyway

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
mickadoo: Instead of using original count I'm using a trait to cleanup the tables before the test. I still don't think this should be done on each test run because it's not relevant to most of the other tests.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Test/Fabricator/Document.php
davialexandre: If we're just comparing a single insert with a single select, then you're right and the performance of the select will probably be faster. Unfortunately, with your change, there will be one insert and one select (The insert has to happen before the call to the `fabricate()`, to create the contact that will be returned by the `get` call). Before this, there was only the insert call. Of course this all depends on how the fabricator will be used in the end. If the same contact is used for all the documents, then your version will probably work better. If not, then the previous would be better.

I really would prefer to not have all this details and performance differences hidden inside the fabricator. Not setting default values for these IDs would allows us to do things like this:

```php
// If I want different contacts for each document
$contact1 = Contact::fabricate();
$document1 = Document::fabricate([
  //other fields
  contact_id => $contact['id'],
  assigned_id => $contact['id'],
 //...
]);

$contact2 = Contact::fabricate();
$document2 = Document::fabricate([
  //other fields
  contact_id => $contact2['id'],
  assigned_id => $contact2['id'],
 //...
]);

// If I want the same contact for multiple documents
$singleContact = Contact::fabricate();
$contactParams = [
  contact_id => $contact2['id'],
  assigned_id => $contact2['id'],
];
$document1 = Document::fabricate(array_merge($contactParams, [
//document specific fields
]));

$document2 = Document::fabricate(array_merge($contactParams, [
//document specific fields
]));

// OR, if you really don't care about the contact
// 1 - disable FK check
// 2 - don't create any contact
$document = Document::fabricate([
  contact_id => 1,
  assignedd_id => 1,
  ...
]);
// enable FK check
```

The last example is used in many places in L&A to avoid us having to create entities that are really not necessary for the test. Check that with your solution, it would not be possible to do something like that, because it would try to find a contact in the database anyway.

mickadoo: I removed all FK references in the defaults of the document fabricator

PR #295, opened 2018-01-12
PCHR-3058: Fix tasks list issues

reneolivo:
## Overview
This PR fixes the following issues with Tasks in the contact section and calendar:

* When clicking "Show More" all tasks would expand or collapse at the same time.
* When Clicking on "Resolved" a "No results found." would always be displayed regardless of the tasks displayed.

## Before
anim
* Note: the resolved panel stops working after opening it for the first time.

## After
anim

## Technical Details

To fix the "Show More" issue the `list.isCollapsed` was changed to a local variable per task: `isCollapsed`. This avoids collapsing all tasks at the same time.

To fix the Resolved tasks the proper variable name was `listResolved` instead of `list.listResolved`:

```html
<div class="alert alert-danger" ng-show="list.listResolvedLoaded && !listResolved.length">No results found.</div>
```

## Comments
BackstopJS tests were run successfully

---

- [x] Tests Pass

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/views/contact/tasks.html
+++ b/uk.co.compucorp.civicrm.tasksassignments/views/contact/tasks.html
@@ -40,7 +40,7 @@
       Loading resolved tasks...
       <div class="{{prefix}}spinner-cover spinner"></div>
     </div>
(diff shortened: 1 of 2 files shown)


reneolivo:
Closed because the branch name is wrong

PR #297, opened 2018-01-22
PCHR-3205: Fix Tests

mickadoo:
## Overview

Since T&A tests started to depend on code from CiviHR they don't work. This PR will get all tests passing and add some new fabricators that CiviHR can use to create Tasks and Documents.

## Before

Tests would not run, and would show the error:

```
Fatal error: Class 'CRM_HRCase_Upgrader' not found in uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php on line 24
``` 

## After

- All tests pass
- There are new fabricators for Task, TaskType and CaseType
- Existing fabricators have been changed to not automatically create related entities, instead requiring the IDs in the `$params` passed to the `fabricate()` method
- We use the `BaseHeadlessTest` instead of the `CiviUnitTestCase`
- Some warnings about missing array indices were causing tests to fail, using the array utils class avoids this

## Notes

- Some other cleanup was done, such as reducing line length
- The contact fabricator was copied over from CiviHR (since we can't depend on CiviHR code)

## When Was It Introduced

After running `git bisect` from the [last time I fixed the T&A tests](https://github.com/compucorp/civihr-tasks-assignments/pull/193) up to staging it pointed to [this commit](https://github.com/compucorp/civihr-tasks-assignments/commit/d09f0ffbdb10f9156dd4634ddf8748c080f0710a) which added the `HRCaseUpgrader` as well as using other code from CiviHR core (`CRM_HRCore_Test_Fabricator_Contact`). 

This might work if you're using your local DB for testing with T&A and CiviHR installed, but when CiviHR is not installed it will fail.

---

- [x] Tests Pass


uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/bootstrap.php
```diff
@@ -9,8 +9,17 @@
 require_once DRUPAL_ROOT . '/includes/bootstrap.inc';
 drupal_bootstrap(DRUPAL_BOOTSTRAP_FULL);
 
+addCiviHRExtensionToIncludePath('/uk.co.compucorp.civicrm.hrcore/');
```
tunbola: Is this right? I thought the extension is supposed to be installed if needed rather than including its files directly?

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
```diff
@@ -15,19 +17,15 @@ class api_v3_DocumentTest extends CiviUnitTestCase {
...
    */
-  public function setUp() {
-    parent::setUp();
-    $this->quickCleanup(['civicrm_activity', 'civicrm_case', 'civicrm_case_activity'], TRUE);
-    $upgrader = CRM_Tasksassignments_Upgrader::instance();
-    $upgrader->install();
-
-    $hrCaseUpgrader = CRM_HRCase_Upgrader::instance();
-    $hrCaseUpgrader->install();
+  public function setUpHeadless() {
```
tunbola: Do we still need this since the test class is extending BaseHeadlessTest?

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
```diff
@@ -35,4 +24,26 @@ public function testDatesForDeletedContactsNotInResult() {
...
   }
+
+  /**
+   * Creates a sample contact
+   *
+   * @param array $params
+   *
+   * @return array
+   */
+  public static function fabricateContact($params = []) {
```
tunbola: Should this function not be in a central place where other T&A tests can use it?

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
```diff
@@ -432,4 +441,19 @@ private function getCustomFieldColumnName($fieldName) {
 
     return CRM_Utils_Array::value($fieldName, $fieldNames);
   }
+
+  /**
+   * @return array
+   */
+  private function fabricateDocumentType() {
```
tunbola: Should this function be in its own class so that other tests can use it?

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/CRM/Tasksassignments/BAO/KeyDates.php
mickadoo: Should be gone now and replaced by `ContactFabricator`

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
mickadoo: Was removed

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/bootstrap.php
mickadoo: Yeah, this was removed since then

uk.co.compucorp.civicrm.tasksassignments/tests/phpunit/api/v3/DocumentTest.php
mickadoo: Good idea, I removed the TaskType fabricator and added a OptionValue fabricator with some convenience methods for document type and task type instead.

PR #302, opened 2018-01-25
PCHR-2567: Fix documents file URLs

reneolivo:
## Overview
This PR fixes two issues with contact's documents files: 

1. New files added to the document wouldn't be downloaded until the page was refreshed.
2. Old files would still be downloaded after removing them from the document.

### Before
anim

### After
anim

### Technical Details

The issue happened because when documents attach or remove files the document ID changes, and this is needed to create the file URL, but this URL was created once and never updated. A workaround was refreshing the page and trying to download the document files again.

To fix this a `$scope.$watch` was added for the document.id property and the file URL updates every time the id changes:

```js
$scope.$watch('document.id', function () {
  $scope.fileUrl = config.url.FILE + '/zip?entityID=' + $scope.document.id + '&entityTable=civicrm_activity';
});
```

Some cosmetic changes like changing the variable name from `urlFile` to `fileUrl` and refactoring the HTML element attributes were also applied.

### Before
anim
The zip file contains 3 files instead of 2.

### After
anim
The zip file contains 2 files as it should.

### Technical Details
The issue happened because zip files were being modified instead of being overwritten. To fix this the following PHP line was modified in CRM/Tasksassignments/Page/Files.php:

```php
// ...
$zip = self::_createZipArchive($zipfullpath);
// ...

static protected function _createZipArchive($zipfullpath) {
  if (file_exists($zipfullpath)) {
    unlink($zipfullpath); // removes the zip file if it already existed.
  }

  $zip = new ZipArchive();
  $zip->open($zipfullpath, ZipArchive::CREATE);

  return $zip;
}
```
**Important:** ZipArchive::OVERWRITE doesn't seem to create new files, just overwrite them so it should not be used. Please see [this ticket](https://compucorp.atlassian.net/browse/PCHR-2567) for more information.

---

- [x] Tests Pass


igorpavlov:
@reneolivo "A workaround was refreshing the page and trying to download the document files again." please put this into technical description instead.

uk.co.compucorp.civicrm.tasksassignments/views/contact/documents.html
```diff
@@ -73,7 +73,10 @@
 
 <script type="text/ng-template" id="document.html">
   <td>
-    <a ng-href="{{fileUrl}}" ng-click="(!document.file_count || document.file_count == '0') && modalDocument('edit', document, $event);" ng-bind="cache.documentType.obj[document.activity_type_id]"></a>
+    <a ng-href="{{fileUrl}}"
+      ng-click="(!document.file_count || document.file_count == '0') && modalDocument('edit', document, $event);">
+      {{cache.documentType.obj[document.activity_type_id]}}
```
igorpavlov: This is different from ng-bind, are you sure this will not produce unexpected behaviour?

uk.co.compucorp.civicrm.tasksassignments/views/contact/documents.html
```diff
@@ -101,7 +104,11 @@
       </ul>
     </div>
   </td>
-  <td><a ng-href="{{fileUrl}}" ng-if="+document.file_count"><i class="fa fa-download"></i></a></td>
+  <td>
+    <a ng-href="{{fileUrl}}" ng-if="+document.file_count">
+      <i class="fa fa-download"></i>
```
igorpavlov: Can we doublecheck it does not produce extra spaces between tags?

uk.co.compucorp.civicrm.tasksassignments/views/dashboard/documents.html
```diff
@@ -182,7 +182,12 @@ <h3 class="panel-title">&nbsp;</h3>
   </div>
 </div>
 <script type="text/ng-template" id="document.html">
-  <td><a ng-href="{{fileUrl}}" ng-click="(!document.file_count || document.file_count == '0') && modalDocument('edit', document, $event);" ng-bind="cache.documentType.obj[document.activity_type_id]"></a></td>
+  <td>
+    <a ng-href="{{fileUrl}}"
+      ng-click="(!document.file_count || document.file_count == '0') && modalDocument('edit', document, $event);">
+      {{cache.documentType.obj[document.activity_type_id]}}
+    </a>
+    </td>
```
igorpavlov: Indentation is wrong

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document.controller.spec.js
```diff
@@ -0,0 +1,41 @@
+/* eslint-env amd, jasmine */
```
igorpavlov: It only tests that the controller is there and if it compiles, but no other tests. What was the initial plan?

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document.controller.spec.js
```diff
@@ -38,20 +40,34 @@ define([
       it('it initializes the document\'s file URL', function () {
         expect($scope.fileUrl).toBe(expectedUrl);
       });
+
+      describe('when the document id is updated', function () {
+        beforeEach(function () {
+          $scope.document.id += 1;
```
igorpavlov: is document.id String or Number?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/document.controller.js
```diff
@@ -20,6 +20,9 @@ define([
       initWatchers();
     }());
 
+    /**
+     * initializes the document watch functions.
```
igorpavlov: Please capitalise and also explain what exactly these watchers do. Ex:
```js
* Initializes the document watch functions.
* - reacts to the document ID change and builds URL for file download.
* - does anything else...
```

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document.controller.spec.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    describe('File URL', function () {
+      var expectedUrl;
+
+      beforeEach(function () {
+        expectedUrl = getDocumentFileUrl();
+      });
+
+      it('it initializes the document\'s file URL', function () {
```
igorpavlov: no need in "it":

it('initializes the document\'s file URL', function () {

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document.controller.spec.js
reneolivo: Oh, sorry, I deleted one of them tests when I force pushed!!!

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document.controller.spec.js
reneolivo: It's a string, but that doesn't matter, it changes regardless of whether it's a string or a number and $scope.watch detects that.

uk.co.compucorp.civicrm.tasksassignments/views/contact/documents.html
reneolivo: How is it different? As far as I know this is the same. Output also looks the same as in the previous version:
civihr_staff compucorp co uk hr17 9000

The reason for moving it it's because it's more explicit to place the element content inside of it using interpolation instead of using `ng-bind`.

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document.controller.spec.js
igorpavlov: right. apologies.

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/document.controller.spec.js
igorpavlov: Not really. If id is "1" I expect it to be "2" according to the code. However it will be "11". If I then compare IDs that could be an issue. Please change to something like:

$scope.document.id = (+$scope.document.id + 1).toString();

uk.co.compucorp.civicrm.tasksassignments/views/contact/documents.html
igorpavlov: Right. However, they are not the same. Please see https://stackoverflow.com/questions/16125872/angularjs-why-ng-bind-is-better-than-in-angular and find an answer where it starts with the word "Visibility:"

PR #309, opened 2018-02-06
PCHR-3145: Automatically select workflow type when param is provided

AkA84:
## Overview
This PR introduces the ability of the T&A app to automatically open the "workflow" modal with a specific assignment type pre-selected. The type id can be passed as part of the query string

open-modal

### Default data of the `modalAssignment` function
Despite having the follow signature
```js
function modalAssignment (data) { }
```
which would be interpreted as the ability of the assignment modal to accept default data, it turned out that the `data` parameter was not used at all by the modal, so any default value would simply be ignored

Adding a simple `angular.copy` call (to follow what the other modal controllers do) in the `init` block allows now to pass default data to the modal

```js
(function init () {
  angular.copy(data, $scope.assignment);

  initWatchers();
}());
```
### Race condition with cache on `$rootScope`
When the assignment type is selected by the user via the select dropdown, the `setData()` function is automatically called, which makes possible to display the list of tasks and documents associated with that given type
```html
<select name="assignment" g-change="setData()" ng-model="assignment.case_type_id" ng-options="value.id as value.title for value in cache.assignmentType.arr">
```
in order to do that, `setData` needs to access the cached data that is stored in the `$rootScope`
```js
function setData () {
  var assignmentType = $rootScope.cache.assignmentType.obj[$scope.assignment.case_type_id];
  // ...
 $scope.assignment.subject = 
 $rootScope.cache.assignmentType.obj[$scope.assignment.case_type_id].title;
```
and in turn the data that is put in the `$rootScope.cache` object is fetched *asynchronously* during the app's `run` block
```js
function run (/* .. */) {
  assignmentService.getTypes().then(function (types) {
    angular.extend($rootScope.cache.assignmentType.obj, types);
    angular.forEach(types, function (type) {
      this.push(type);
    }, $rootScope.cache.assignmentType.arr); // <--
  });
}
```
The fact that the data was fetched asynchronously and that the modal didn't wait to make sure that it was actually fetched before calling `setData` wasn't a problem when the only way to call `setData` was via user interaction, because by the time the user could interact with the select dropdown, the data was surely already fetched

The problem was that now `setData` could be called on `init` if the case type id was set automatically, and this lead to a race condition where the function would be called before the cached data could be loaded

The solution was a simple watcher that would wait until the data was ready, before calling the function
```js
function initWatchers () {
  var listener = $rootScope.$watch('cache.assignmentType.obj', function (cache) {
    if (!_.isEmpty(cache)) {
      setData();
      listener();
    }
  }, true);
}
```
if the cache type id was passed, then the data would be set automatically, otherwise nothing would happen (there is a check in place already in the function)
```js
function setData () {
  if (!assignmentType) {
    $scope.activity.activitySet = {};
    $scope.assignment.subject = '';

    return;
  }
}
```

- [x] Tests Pass


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/main.controller.js
```diff
@@ -35,7 +37,7 @@ define([
       var method = $rootScope['modal' + _.capitalize(modalType)];
 
       if (typeof method === 'function') {
-        method();
+        method({ case_type_id: (queryParams.caseTypeId || null) });
```
reneolivo: Isn't it confusing to pass `{ case_type_id: null }` to all modals other than modalDocument?

Why not do:

```js
function modalDocument (data, modalMode, e) {
  // ...
  data = data || {};
  data.case_type_id = data.case_type_id || queryParams.caseTypeId || null;
  // ...
}
```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/modal/modal-assignment.controller.js
```diff
@@ -301,6 +303,13 @@ define([
 
         angular.copy(documentList, $scope.documentList);
       });
+
+      var listener = $rootScope.$watch('cache.assignmentType.obj', function (cache) {
```
reneolivo:
```js
function initWatchers () {
  var rootScopeWatchUnlisten; // or a better name, but define it at the top.
  // ...
  rootScopeWatchUnlisten = $rootScope.$watch('cache.assignmentType.obj', function (cache) {
    // ...
    rootScopeWatchUnlisten();
    // ...
  }, true);
}
```

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-assignments/controllers/main.controller.js
AkA84: the modal is modalAssignment, to be clear.

I initially did what you proposed but it doesn't work correctly, as it would use the query params even when you manually open the modal (by clicking on the "add workflow" button). You want the case type to be auto-selected only when the modal is opened automatically.

And yes i wasn't much too happy with passing case_type_id to all the modal types, but i don't know how to pass it only to the assignment modal without making a ad-hoc condition.

So I thought that in the end it shouldn't be a problem: if the modal uses that value, fine, otherwise it will just ignore it

PR #310, opened 2018-02-07
PCHR-3105: Run T&A tests with Jenkins

davialexandre:
## Overview

This PRs adds the necessary changes in order for Jenkins to run the T&A tests.

## Before

Jenkins couldn't run the extension's tests

## After

Jenkins can run the extension's tests. The same way as it works for CiviHR, it will run the tests in these specific scenarios:

- a PR has the `ready to test` label
- a merge is merged to `staging` or `master`.

Notifications are sent when a build starts, fails and/or finishes. Since we don't have a specific room for the T&A project on hipchat, they will be sent to the CiviHR room.

## Technical Details

The Jenkinsfile added here was heavily based on the [CiviHR one](https://github.com/civicrm/civihr/blob/staging/Jenkinsfile). However, since we have a single extension to be tested here, all the code related to looping through a list of extensions was removed.

T&A is supposed to be a standalone extension and to not depend on anything from CiviHR. However, as it's possible to see [here](https://github.com/compucorp/civihr-tasks-assignments/blob/767caf48b53cfee4a3140d09e64f8704f1dde52d/uk.co.compucorp.civicrm.tasksassignments/js/karma.conf.js#L31) it depends on reqangular, an generic utility library extension that, currently, is in the CiviHR repo. For that reason, in order to run the tests, we need to clone the civihr repo and install reqangular (but not the other CiviHR extensions).

While working on this, I also noticed that the extension had a dependency on the [SSP](https://github.com/compucorp/civihr-employee-portal) module. There were two places using the [_get_uf_match_contact()](https://github.com/compucorp/civihr-employee-portal/blob/staging/civihr_employee_portal/civihr_employee_portal.module#L1755-L1776) function. To remove the dependency, I simply copied the function from the SSP to the place where it is used by the extension. This was done on 767caf4.

Another issue I found while working on this was some hardcoded paths in the karma configuration file. I replace them with calls to `cv()`. This was done in cc4b766. Unfortunately, I was not able to remove all the hardcoded paths in JS files. There is still one ocurrence of that in the [globals.js file](https://github.com/compucorp/civihr-tasks-assignments/blob/staging/uk.co.compucorp.civicrm.tasksassignments/js/test/globals.js#L10). Since this is not a file loaded by node (it's injected by karma into the browser used to run the JS tests), we cannot use the `cv()` command there and I couldn't find any easy way to pass a dynamic file path to it. This means that, currently, to run the JS tests, T&A needs to always be installed inside that hardcoded path. Otherwise it won't be able to load the necessary files.

Finally, to be able to run the JS tests with Jenkins, the karma file had to be updated to use Chrome in Headless mode. This was done in fc82fc7

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
@@ -943,7 +943,7 @@ class CRM_Tasksassignments_Reminder {
   /**
    * Loads civicrm_uf_match data based on passed contact_id
    *
-   * @param $contact_id
+   * @param int $contact_id
    *
    * @return mixed
    *
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
@@ -638,8 +638,7 @@ class CRM_Tasksassignments_Reminder {
   public static function createActivityURL($contactId = NULL, $activityResultId = NULL) {
 
     if ($contactId) {
-      $activityUrl = '';
-      $drupal_user = user_load(_get_uf_match_contact($contactId)['uf_id']);
+      $drupal_user = user_load(self::getUfMatchContact($contactId)['uf_id']);
 
       if (user_access('access CiviCRM', $drupal_user)) {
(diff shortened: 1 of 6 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
```diff
@@ -941,4 +940,27 @@ private static function _send($contactId, $email, $body_subject, $body_html) {
     return $result;
   }
 
+  /**
+   * Loads civicrm_uf_match data based on passed contact_id
+   *
+   * @param $contact_id
```
tunbola: Missing param type

tunbola:
typo in PR description s/runt/run

PR #313, opened 2018-02-14
PCHR-3284: Select current user as target contact in workflow modal

AkA84:
## Overview
This PR fixes a regression issues introduced in #309 , where the current user would not be automatically selected as the target contact in the workflow modal anymore

## Before
before

## After
after

## Technical Details
The reason for the bug was this line
```js
angular.copy(data, $scope.assignment);
```
introduced in `ModalAssignmentController` in this [commit](https://github.com/compucorp/civihr-tasks-assignments/pull/309/commits/2f1a7ff4e66038d1352e9822df9ed519af88f670#diff-7e24d6add6cfdc8f257b7ddc9f6cfadaR74)

In case a `destination` object (in this case `$scope.assignment`) is given to `angular.copy`, then

> all of its elements (for arrays) or properties (for objects) are deleted and then all elements/properties from the source are copied to it.

_(from [angular.copy](https://docs.angularjs.org/api/ng/function/angular.copy))_

so basically all the properties defined on `$scope.assignment` before that line
```js
$scope.assignment = {};
$scope.assignment.status_id = '1';
$scope.assignment.contact_id = config.CONTACT_ID;
$scope.assignment.client_id = $scope.assignment.contact_id;
$scope.assignment.subject = '';
$scope.assignment.dueDate = null;
```
would get erased by `data`, which is either empty (most of the cases) or it simply has only the `case_type_id` property set

A simple solution to the problem is to do the copy immediately on `$scope.assignment` declaration, instead of waiting for `init()` to be called
```js
$scope.assignment = angular.copy(data);
$scope.assignment.status_id = '1';
$scope.assignment.contact_id = config.CONTACT_ID;
$scope.assignment.client_id = $scope.assignment.contact_id;
$scope.assignment.subject = '';
$scope.assignment.dueDate = null;
```


uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-assignment.controller.spec.js
```diff
@@ -34,6 +34,19 @@ define([
       angular.extend(scope, Mock);
     }));
 
+    describe('on init', function () {
+      var resolvedData = { foo: 'foo', bar: 'bar' };
```
igorpavlov: The 'bar' property is not used in tests. It either should be used or be deleted.

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-assignment.controller.spec.js
```diff
@@ -34,6 +34,19 @@ define([
...
+
+      beforeEach(function () {
+        initController(resolvedData);
```
igorpavlov: Please do deepClone here. So then you don't need to care about mutation and also about testing.

uk.co.compucorp.civicrm.tasksassignments/js/test/controllers/modal/modal-assignment.controller.spec.js
```diff
@@ -34,6 +34,19 @@ define([
...
+      });
+
+      it('copies the passed data in the $scope assignment object', function () {
+        expect(scope.assignment.foo).toBeDefined();
```
igorpavlov: If you deepClone object (as described above), you don't need to care if it was defined.

PR #317, opened 2018-02-26
PCHR-3353: Set Custom Groups to Reserved

ajesamson:
## Overview
This PR sets custom field groups that should not be editable to "is_reserved"=yes for the following custom groups.

- Activity Custom Fields
- Probation

## Before
custom_group_before

## After
custom_group_after_2

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -712,6 +712,21 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
     return TRUE;
   }
 
+    /**
+     * Upgrade CustomGroup, setting Probation and Activity_Custom_Fields to is_reserved Yes
+     * @return bool
+     */
+  public function upgrade_1032() {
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -712,6 +712,22 @@ public function upgrade_1031() {
...
+  /**
+   * Upgrade CustomGroup, setting Probation and Activity_Custom_Fields to is_reserved Yes
+   *
+   * @return bool
+   */
+  public function upgrade_1032() {
+    civicrm_api3('CustomGroup', 'get', [
+      'sequential' => 1,
+      'return' => ['id'],
+      'name' => ['IN' => ['Probation', 'Activity_Custom_Fields']],
```
mickadoo: The PR description only mentions "Activity_Custom_Fields", which is correct?

ajesamson: This PR only cater for 'Activity_Custom_Fields'. Other changes are in a separate PR. That is they are excluded. Same with the other PR, Activity_Custom_Fields was excluded.

tunbola:
@ajesamson, looks like you need to rebase your commit messages and prefix them with the ticket number.

PR #322, opened 2018-03-05
PCHR-3403: Allow only the "Add timeline" action

reneolivo:
## Overview
This PR fixes the following issues:

* Replaces the actions dropdown with a single button for the "Add timeline" action.
* Moves the help text outside of the case type form.

## Before
new workflow type hr17 9000 3

## After
new workflow type hr17 9000 9

### Add timeline action

The CiviCRM's alter angular hook was implemented in order to replace the dropdown:

```php
function tasksAssignments_civicrm_alterAngular(\Civi\Angular\Manager $angular) {
  $changeSet = \Civi\Angular\ChangeSet::create('Workflow Modifications');

  $changeSet->alterHtml('~/crmCaseType/edit.html', function (phpQueryObject $doc) {
    _tasksAssignments_change_workflow_help_text($doc);
    _tasksAssignments_remove_non_civihr_tabs_from_workflow($doc);
    _tasksAssignments_allow_only_add_timeline_action($doc); // <--- here
  });

  $angular->add($changeSet);
}

function _tasksAssignments_allow_only_add_timeline_action (phpQueryObject $doc) {
  $actionDropdown = $doc->find('select[ng-model="newActivitySetWorkflow"]');
  $addTimelineBtn = '
    <button class="btn btn-secondary pull-right"
      ng-show="isNewActivitySetAllowed(\'timeline\')"
      ng-click="addActivitySet(\'timeline\')">
      Add timeline
    </button>';

  $actionDropdown->after($addTimelineBtn); // adds the new button
  $actionDropdown->remove(); // removes the previous action dropdown
}
```

the `ng-show="isNewActivitySetAllowed('timeline')"` is there because:

1. The previous actions dropdown had it defined for the add timeline and add sequence actions ([here](https://github.com/civicrm/civicrm-core/blob/master/ang/crmCaseType/edit.html#L34)).
2. Because this CiviCRM function can determine if adding a new timeline is allowed, although for now [it always returns true](https://github.com/civicrm/civicrm-core/blob/master/ang/crmCaseType.js#L281) this might change in the future:

```js
$scope.isNewActivitySetAllowed = function(workflow) {
  switch (workflow) {
    case 'timeline':
      return true; // <--- always returns true for the timeline.
    case 'sequence':
      return 0 === _.where($scope.caseType.definition.activitySets, {sequence: '1'}).length;
    default:
      CRM.console('warn', 'Denied access to unrecognized workflow: (' + workflow + ')');
      return false;
  }
};
```

### Moving the help text

The `change_workflow_help_text` function was modified so it moves the help text outside of the case type form:

```php
function _tasksAssignments_change_workflow_help_text(phpQueryObject $doc) {
  $helpBlock = $doc->find('.crmCaseType .help');

  $text = '<p>...</p>';
  $text .= '<p>...</p>';

  $helpBlock->html($text);
  // Places the help text outside of the case type form:
  $helpBlock->insertBefore('.crmCaseType'); // <---- here
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
@@ -438,6 +438,8 @@ function tasksassignments_extensionsPageRedirect()  {
  * @param phpQueryObject $doc
  */
 function _tasksAssignments_change_workflow_help_text(phpQueryObject $doc) {
+  $helpBlock = $doc->find('.crmCaseType .help');
+
   $text = '<p>' . ts('Configure your workflow timelines below. Each workflow type can have several
     different task timelines. Each timeline allows you to set different tasks and documents which
     become part of your task list on your task dashboard. As such different timelines can be setup
@@ -448,7 +450,9 @@ function _tasksAssignments_change_workflow_help_text(phpQueryObject $doc) {
     be used for other processes too, such as a person going on maternity leave or moving region or
     location.') . '</p>';
 
-  $doc->find('.crmCaseType .help')->html($text);
+  $helpBlock->html($text);
+  // Places the help text outside of the case type form:
+  $helpBlock->insertBefore('.crmCaseType');
 }
 
 /**
--- a/uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
@@ -284,6 +284,7 @@ function tasksAssignments_civicrm_alterAngular(\Civi\Angular\Manager $angular) {
   $changeSet->alterHtml('~/crmCaseType/edit.html', function (phpQueryObject $doc) {
     _tasksAssignments_change_workflow_help_text($doc);
     _tasksAssignments_remove_non_civihr_tabs_from_workflow($doc);
+    _tasksAssignments_allow_only_add_timeline_action($doc);
   });
 
   $angular->add($changeSet);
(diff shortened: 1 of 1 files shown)


JonScreat:
@reneolivo can you confirm for the the font and font-weight used in the button please? Looks a little light (hard to tell from just a screenshot). It should be open sans semibold.

Also there are a couple of issues elsewhere on this screen with the borders. Will they be addressed elsewhere?

reneolivo:
@JonScreat it seems jQuery UI is overriding the style for buttons inside of widgets so they all have the Verdana font. So far I think that's the only change it does.

Should I override the font for all buttons inside jQuery UI buttons or just this one?

JonScreat:
@reneolivo for all buttons please. The only font we should be using is Open Sans. Thanks

jamienovick:
@reneolivo @JonScreat @mukeshcompucorp why are we even loading Verdana at all? @AkA84 @mukeshcompucorp is this because Seven is the Drupal base theme we are using? Is the plan to remove that theme and create our own blank Drupal theme for the purpose of CiviHR admin?

reneolivo:
@jamienovick the Verdana font rule comes from jQuery UI's theme (you can see the definition [here](https://github.com/components/jqueryui/blob/master/themes/smoothness/jquery-ui.css#L896)) and jQuery UI is in turn loaded by CiviCRM.

This rule only affects buttons inside of jQuery UI widgets so it's very rare to spot them. I think we found this issue by chance.

mukeshcompucorp:
@reneolivo Can you move that green help section outside form block as well? by standard it should set on outside to look something like
image

cc @JonScreat

JonScreat:
@reneolivo additional comments:
image

- panel should have 2px border radius
- save/cancel buttons should be to the left of the panel
- There are issues with borders doubling up and making the lines to think (see the line underneath the tab)

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -463,3 +468,22 @@ function _tasksAssignments_remove_non_civihr_tabs_from_workflow (phpQueryObject
...
 }
+
+/**
+ * Removes the Workflow configuration's actions dropdown and replaces it with a
+ * button that only allows the "Add timeline" action.
+ *
+ * @param phpQueryObject $doc
+ */
+function _tasksAssignments_allow_only_add_timeline_action (phpQueryObject $doc) {
+  $actionDropdown = $doc->find('select[ng-model="newActivitySetWorkflow"]');
```
tunbola: Was there any reason why you did not go the approach of probably removing the unwanted option values in the select dropdown rather than creating a new button element.
I'm of the opinion that the latter would require minimal changes should anything change in the core code and also ensure that the html structure from core is preserved.
What do you think?

reneolivo: @tunbola having a single "Add timeline" button was part of the requirement ([wiki link](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/134906233/Tasks+and+Workflows+-+Administrator+UX+improvements)), but even then, I think removing the dropdown is the way to go. 

If we only remove the options that we don't need it will leave only a single option in the dropdown (even if that might change in the future) so to allow users to add timelines we'd either need to add some sort of "run selected action" button next to the dropdown or add an empty "Please select an action" button.

Since we'll only have one available action for now it makes sense to turn it into a button.

PR #324, opened 2018-03-09
PCHR-3382: Send Daily reminder Email to Recipient With Active Contract and With Primary Work Email

tunbola:
## Overview
Currently the daily reminder emails go to recipient without active contracts and to emails that are not the primary work emails of the recipients. This PR fixes this issue and makes sure that only recipients with active contracts get daily reminder emails and such emails goes to the primary work email of the recipient, if both criteria are not met, i.e recipient has an active contract and has a primary work email, the email us not sent to such a recipient even if it satisfies other criteria.

## Before
- The daily reminder email goes to recipient without active contracts and to emails that are not the primary work emails of the recipients.
This is due to the fact that the two queries for getting recipients that is merged together via a [UNION](https://github.com/compucorp/civihr-tasks-assignments/blob/5053e7f732d12c4b678bc5a086bf53f65e6e2cd1/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php#L267) [Query 1](https://github.com/compucorp/civihr-tasks-assignments/blob/5053e7f732d12c4b678bc5a086bf53f65e6e2cd1/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php#L428) and [Query 2](https://github.com/compucorp/civihr-tasks-assignments/blob/5053e7f732d12c4b678bc5a086bf53f65e6e2cd1/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php#L480) are inconsistent with the email types to filter by, one filters by where ` AND e.is_primary = 1` i.e where the email is primary while the other does not have this filter at all.
- Also there is no check for whether the recipient has an active contract or not before the email is sent.

## After
- The two queries for getting the recipients mentioned in the after section were modified to filter by the primary work email of the contact. 
- The `HrJobcontract.getcontactswithcontractsinperiod` API was used to filter off contacts without an active contract before the emails were sent to recipients.

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
@@ -9,6 +9,7 @@ class CRM_Tasksassignments_Reminder {
   private static $_activityOptions = array();
   private static $_relatedExtensions = array(
     'appraisals' => false,
+    'hrjobcontract' => false
   );
   private static $_reminderSettings = [];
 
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php
```diff
@@ -299,12 +318,33 @@ public static function sendDailyReminder() {
...
+      $result = civicrm_api3('HRJobContract', 'getcontactswithcontractsinperiod',[
+        'contact_id' => ['IN' => $contactID],
+      ]);
+
+      return array_column($result['values'], 'id');
+    }
+
+    return $contactID;
+  }
+
```
davialexandre: Please add tests to cover these changes

davialexandre: After discussing with @tunbola, it turns out that writing tests to this functionally will be much much more complex than I originally expected. The reason is that it is a highly coupled code, which dozens of dependencies that make the test setup very complex (there is even a piece of code [making queries directly to the Drupal database](https://github.com/compucorp/civihr-tasks-assignments/blob/a658cc58bee4a56984d796d5171fbb752cb6a885/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Reminder.php#L426-L436)!). To be able to test this, a considerable refactoring will be necessary, but that only increases the risk of introducing a new bug (because we don't have tests for the existing functionality). Given all that, I decided to make an exception this and I'm approving without the addition of tests.

PR #325, opened 2018-03-13
PCHR-3376: Update T&A Administer Menu Items

mickadoo:
## Overview

We are changing what menu items can be accessed. T&A provides some items for the "Configure" submenu, and these will be altered.

## Before

image

- T&A provided the "Tasks" submenu in the "Configure" menu but no items had permission set
- It cloned two items from the CiviCase submenu, which were relabled as "Workflows" and "Workflows Status"
- The weight of the "Tasks" item was not set and so it appears at the top of the list

## After

image

- The permission for all T&A menu items is `administer CiviCase`. 
- The "Workflows Status" menu item is no longer cloned. 
- The weight of "Tasks" has been set so it appears in the required order (before "Leave")

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -740,20 +740,23 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
    * @return bool
    */
   public function upgrade_1033() {
-    $params = ['return' => 'id', 'name' => 'tasksassignments_administer'];
-    $parentId = (int) civicrm_api3('Navigation', 'getvalue', $params);
-    $params = ['return' => 'id', 'name' => 'ta_settings'];
-    $taSettingsId = (int) civicrm_api3('Navigation', 'getvalue', $params);
     $permission = 'administer CiviCase';
+    $itemsToChange = ['tasksassignments_administer', 'ta_settings'];
 
-    // Update permissions
-    foreach ([$parentId, $taSettingsId] as $navId) {
-      $params = ['permission' => $permission, 'id' => $navId];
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -734,6 +734,30 @@ public function upgrade_1032() {
...
+  /**
+   * Update permissions and set the weight for the "Tasks" menu item.
+   *
+   * @return bool
+   */
+  public function upgrade_1033() {
+    $params = ['return' => 'id', 'name' => 'tasksassignments_administer'];
+    $parentId = (int) civicrm_api3('Navigation', 'getvalue', $params);
+    $params = ['return' => 'id', 'name' => 'ta_settings'];
+    $taSettingsId = (int) civicrm_api3('Navigation', 'getvalue', $params);
```
tunbola: Why don't we just call the Navigation API once to get the ids of the menu's rather than twice?

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -734,6 +734,30 @@ public function upgrade_1032() {
...
+    $permission = 'administer CiviCase';
+
+    // Update permissions
+    foreach ([$parentId, $taSettingsId] as $navId) {
+      $params = ['permission' => $permission, 'id' => $navId];
+      civicrm_api3('Navigation', 'create', $params);
+    }
+
+    // Update parent weight
+    civicrm_api3('Navigation', 'create', ['id' => $parentId, 'weight' => -97]);
```
tunbola: We can actually update permission and weight of the parentID in one call and the permission of taSettingsId in another call, which means we will call the API two times rather than three.
We can have a function do this that will optionally update the permission or not if present

PR #334, opened 2018-04-09
PCHR-3405: Rename workflow statuses

reneolivo:
## Overview
This PR renames the following workflow statuses:

|From|to|
|----|--|
|Ongoing|In Progress|
|Resolved|Completed|

## Before
case status options hr17 9200 1

## After
case status options hr17 9200

## Technical Details
A new upgrader was added to T&W:

```php
<?php

class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
{
  // ...
  public function upgrade_1032() {
    $this->_renameCaseStatus('Ongoing', 'In Progress');
    $this->_renameCaseStatus('Resolved', 'Completed');
    return TRUE;
  }

  private function _renameCaseStatus($fromLabel, $toLabel) {
    $optionValue = civicrm_api3('OptionValue', 'getsingle', [
      'option_group_id.name' => 'case_status',
      'label' => $fromLabel,
      'is_reserved' => 1,
      'options' => [ 'limit' => 1 ],
    ]);
    civicrm_api3('OptionValue', 'create', [
      'id' => $optionValue['id'],
      'label' => $toLabel
    ]);
  }
}
```

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -714,10 +714,12 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base
 
   /**
    * Renames the case status Ongoing and Resolved to In Progress and Completed.
+   *
+   * @return bool
    */
   public function upgrade_1032() {
-    $this->_renameCaseStatus('Ongoing', 'In Progress');
-    $this->_renameCaseStatus('Resolved', 'Completed');
+    $this->_relabelCaseStatus('Ongoing', 'In Progress');
+    $this->_relabelCaseStatus('Resolved', 'Completed');
 
     return TRUE;
   }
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -712,6 +712,16 @@ public function upgrade_1031() {
     return TRUE;
   }
 
+  /**
+   * Renames the case status Ongoing and Resolved to In Progress and Completed.
+   */
```
mickadoo: missing `@return bool`

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -793,4 +803,24 @@ private function _uninstallActivityTypes($component, array $types) {
...
     }
   }
+
+  /**
+   * Renames the label for a case status
+   *
+   * @param string $fromLabel
+   * @param string $toLabel
+   */
+  private function _renameCaseStatus($fromLabel, $toLabel) {
```
mickadoo: I wonder would this be more accurate if it was called `_relabelCaseStatus`? Since these are option values they also have a `name` field, which 'rename` might sound like it is changing.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -793,4 +803,24 @@ private function _uninstallActivityTypes($component, array $types) {
...
+    $optionValue = civicrm_api3('OptionValue', 'getsingle', [
```
mickadoo: In general we try to make the upgrader functions idempotent, just in case something goes wrong and we need to run them again. I think because you're doing a `getsingle` call here and using the label in the criteria it will fail the second time. Can you rewrite this so that it could be run multiple times, and does a safety check to ensure the `optionValue` exists before the `create` call? 

It might be at some point in the future the base case statuses are changed in a fresh installation, in that case we wouldn't be sure if this status exists.

PR #338, opened 2018-04-17
PCHR-3575: Hide Columns and Show component type

deb1990:
## Overview
As part of this PR the following features are developed
**Columns Hidden**
- Status
- Reference
- Select

**Renamed the column**
- "Activity" to "Task / Document Type"

**Added component type with activity type when displaying in the column**

**Default value of Reference is set to NULL**

## Before
2018-04-18 at 11 56 am

## Technical Details
1. Moved the content of `crm-case-type.config.js` to `route.decorator.js` to better organize.
2. From the above decorator, `CiviTask` and `CiviDocument` activity types are sent in separate objects, which is picked up by `CaseTypeExtendedController` and component type labels are added.
3. `$scope.addActivity` is overridden. It calls the parent function first, then removes the `reference_activity` property.
4. `_tasksAssignments_remove_columns_from_timeline` is added to remove columns.
5. `_tasksAssignments_change_column_text` is added to change the label.


AkA84:
test for slack

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
```diff
@@ -1,44 +1,92 @@
...
+        initController($scope);
+        $scope.addActivity({ activityTypes: activityTypes }, activityType);
+      });
+
+      it('calls the parent function', function () {
+        expect(addActivityInParent).toHaveBeenCalledWith({ activityTypes: activityTypes }, activityType);
+      });
+
+      it('removes the reference property for the activity type', function () {
+        expect(activityTypes[0].reference_activity).toBeUndefined();
```
reneolivo: Did you confirm if removing this property doesn't mess up creating a new case?

The `reference_activity` along with the `offset` are used to determine the date for the activity. Ex: activity "Schedule joining date", reference activity is "Open Case" (first activity), offset is 3, date of the case is "2020-01-01", then the date for the the "Schedule joining date" will be "2020-01-04".

Can you confirm that still works?

* Go to  `/civicrm/tasksassignments/dashboard#/tasks` 
* Create a new workflow
* Check if the dates for each task are set properly.

Thanks.

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
```diff
@@ -1,44 +1,92 @@
...
+      });
+    });
+
+    function initController ($scope) {
```
reneolivo: missing doc

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
```diff
@@ -1,44 +1,92 @@
...
         apiCalls: {},
-        activityOptions: activityOptions
+        activityOptionsTask: { values: angular.copy(TaskActivityOptionsData) },
+        activityOptionsDocument: { values: angular.copy(DocumentActivityOptionsData) }
       });
 
       $scope.$digest();
     }
+
+    function getActivityTypes () {
```
reneolivo: missing doc

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -510,3 +512,37 @@ function _tasksAssignments_allow_only_add_timeline_action (phpQueryObject $doc)
...
   $actionDropdown->remove();
 }
+
+/**
+ * Remove columns from timeline
+ *
+ * @param phpQueryObject $doc
+ */
+function _tasksAssignments_remove_columns_from_timeline ($doc) {
+  $selectColumnHeader = $doc->find('table th:nth-child(6)');
```
reneolivo: Finding things like `table th:nth-child(6)` makes me a little nervous because the order might change, or a new column might be added and this will remove a column that is not intended.

Can we use `$doc->find('table th:contains("Status")` ? There are some benefits to this:

* What you are trying to select is more clear.
* It won't change the outcome if the order of the columns change or a new column is added. It also won't remove a random column if this happens.
* If the "Status" column is renamed, it will be visible again. That's the only drawback.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -510,3 +512,37 @@ function _tasksAssignments_allow_only_add_timeline_action (phpQueryObject $doc)
...
+  $statusColumnHeader = $doc->find('table th:nth-child(3)');
+  $statusColumnBody = $doc->find('table td:nth-child(3)');
+
+  $elementsToRemove = [
+    $selectColumnHeader, $selectColumnBody,
+    $referenceColumnHeader, $referenceColumnBody,
+    $statusColumnHeader, $statusColumnBody
+  ];
+
+  foreach( $elementsToRemove as $element ) {
```
reneolivo: can we refactor this function a bit?

```php
function _tasksAssignments_remove_column_from_timeline ($doc, columnName) {
  $columnHeader = $doc->find('table th:contains("' + columnName + "');
  $column = $doc->find('table td:nth-child(' + $columnHeader->index() + ')');

  $columnHeader->remove();
  $column->remove();
}
```

And use it either like this:
```php
$changeSet->alterHtml('~/crmCaseType/timelineTable.html', function (phpQueryObject $doc) {
  // ...
 _tasksAssignments_remove_column_from_timeline($doc, 'Status');
 _tasksAssignments_remove_column_from_timeline($doc, 'Reference');
 _tasksAssignments_remove_column_from_timeline($doc, 'Select');
});
```

Or like this:
```php
$changeSet->alterHtml('~/crmCaseType/timelineTable.html', function (phpQueryObject $doc) {
  // ...
  $columnsToBeRemovedFromTimeline = [ 'Status', 'Reference', 'Select' ];

  foreach ($columnsToBeRemovedFromTimeline as $columnName) {
    _tasksAssignments_remove_column_from_timeline($doc, $columnName);
  }
});
```

tunbola: The second option seems more preferable, or better still, the initial function name `_tasksAssignments_remove_columns_from_timeline` can be retained while the loop is done within the function for the columns that need to be removed. This will help keeping the hook uncluttered with unnecessary logic.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -510,3 +512,37 @@ function _tasksAssignments_allow_only_add_timeline_action (phpQueryObject $doc)
...
+  $referenceColumnHeader = $doc->find('table th:nth-child(4)');
+  $referenceColumnBody = $doc->find('table td:nth-child(4)');
+  $statusColumnHeader = $doc->find('table th:nth-child(3)');
+  $statusColumnBody = $doc->find('table td:nth-child(3)');
+
+  $elementsToRemove = [
+    $selectColumnHeader, $selectColumnBody,
+    $referenceColumnHeader, $referenceColumnBody,
+    $statusColumnHeader, $statusColumnBody
+  ];
```
tunbola: Please format in such a way that each item stays on its own line.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
tunbola: I agree.

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
deb1990: yeah it works.

2018-04-18 at 7 01 pm

2018-04-18 at 7 02 pm

reneolivo: Thanks for checking

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
deb1990: Code is updated

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -289,8 +289,12 @@ function tasksAssignments_civicrm_alterAngular(\Civi\Angular\Manager $angular) {
 
   $changeSet->alterHtml('~/crmCaseType/timelineTable.html', function (phpQueryObject $doc) {
     _tasksAssignments_change_add_activity_dropdown_placeholder($doc);
-    _tasksAssignments_remove_columns_from_timeline($doc);
     _tasksAssignments_change_column_text($doc);
+
+    $columnsToBeRemovedFromTimeline = [ 'Status', 'Reference', 'Select' ];
+    foreach ($columnsToBeRemovedFromTimeline as $columnName) {
+      _tasksAssignments_remove_column_from_timeline($doc, $columnName);
```
tunbola: Can we move the loop inside the `_tasksAssignments_remove_column_from_timeline` function so as to keep the hook clean and free from a lot of logic and also consistency with other helper methods in the hook? The logic should reside in the function. The function name can be pluralized like you did initially `_tasksAssignments_remove_columns_from_timeline`.

PR #339, opened 2018-04-18
test for slack

AkA84:
test for slack


davialexandre:
I believe that the testing is done. Closing

PR #348, opened 2018-04-23
PCHR-3532: Only display workflows in workflow configuration screen

reneolivo:
## Overview
This PR only displays Workflows case types on the Workflow configuration screen
(`/civicrm/a/#/caseType/`).

## Before
screenshot 2018-04-23 16 39 29
* **Note:** the Application case type is of the Vacancy category and should not be displayed here.

## After
screenshot 2018-04-23 16 37 23
* **Note:** the *Category* column was removed since only Workflow categories are displayed.

## Technical Details

The route decorator introduced in [this PR](https://github.com/compucorp/civihr-tasks-assignments/pull/338) was enhanced so it replaces the case type api call defined in [the original core definition](https://github.com/civicrm/civicrm-core/blob/master/ang/crmCaseType.js#L41):

```js
(function init () {
  var caseTypeListRoute = $delegate.routes['/caseType'];

  // ...

  caseTypeListRoute.resolve.caseTypes = getResolverForWorkflowTypes();
})();

function getResolverForWorkflowTypes () {
  workflowTypesResolver.$inject = ['crmApi'];

  function workflowTypesResolver (crmApi) {
    return crmApi('CaseType', 'get', {
      category: 'WORKFLOW', // <--- API Call replacement: only brings workflow case types
      options: { limit: 0 }
    });
  }

  return workflowTypesResolver;
}
```

## Comments

* The init statements defined in the `caseType.extended.controller` file were moved to a IIFE so they follow the convention.
* The Category column was removed from the workflow configuration list since only case type of the Workflow category would be displayed and made the column redundant.
* `civilint` complained about the hooks' documentation format. Since parameters and return values of a hook are supposed to be documented in the original docs they were removed.


uk.co.compucorp.civicrm.tasksassignments/js/src/crm-tasks-workflows/controllers/casetype.extended.controller.js
```diff
@@ -26,7 +30,7 @@
      */
     function addActivity (activitySet, activityType) {
       // call parent
```
igorpavlov: This comment is excessive, the code now is self-explanatory. Please delete this comment.

uk.co.compucorp.civicrm.tasksassignments/js/src/crm-tasks-workflows/decorators/route.decorator.js
```diff
@@ -34,6 +43,24 @@
...
 
+    /**
+     * Returns a resolver for the list of workflow case types.
+     *
+     * @return {Function} the function that will resolve the workflow case types.
+     */
+    function getResolverForWorkflowTypes () {
+      workflowTypesResolver.$inject = ['crmApi'];
+
+      function workflowTypesResolver (crmApi) {
```
igorpavlov:
```js
var workflowTypesResolver = function (crmApi)
  ...
}
```
That would be probably easier to read, you define a var, then return it. It will also ensure the order of code execution.

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/decorators/route.decorator.spec.js
```diff
@@ -0,0 +1,81 @@
+/* eslint-env jasmine */
+
+(function () {
+  'use strict';
+
+  describe('RouteDecorator', function () {
+    var $route, crmApiSpy;
+
+    beforeEach(module('ngRoute', 'crm-tasks-workflows.decorators', function ($routeProvider) {
+      // This happens before the configuration phase:
```
igorpavlov: The comment would be much more valuable if "this happens" is replaced with an actual explanation on what's going on.

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/decorators/route.decorator.spec.js
```diff
@@ -0,0 +1,81 @@
...
+      $routeProvider.when('/caseType/:id', {
```
igorpavlov:
```js
$routeProvider
  .when('/caseType/:id', {... })
  .when('/caseType', {... })
```
as per Styleguide

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/decorators/route.decorator.spec.js
igorpavlov: We also should leave a comment on why we have empty funcs in the code and what's happening to these funcs afterwards. Imho, so up to you.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -15,10 +15,7 @@ function tasksassignments_civicrm_config(&$config) {
 }
 
 /**
- * Implements hook_apiWrappers.
- *
- * @param array $wrappers
- * @param array $apiRequest
+ * Implements hook_apiWrappers().
```
tunbola: Why remove the parameter doc block?

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -39,9 +36,7 @@ function tasksassignments_civicrm_apiWrappers(&$wrappers, $apiRequest) {
 }
 
 /**
- * Implementation of hook_civicrm_xmlMenu
- *
- * @param $files array(string)
+ * Implements hook_civicrm_xmlMenu().
```
tunbola: Please add the param doc block back.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -59,57 +54,51 @@ function tasksassignments_civicrm_install() {
...
 
 /**
- * Implementation of hook_civicrm_upgrade
- *
- * @param $op string, the type of operation being performed; 'check' or 'enqueue'
- * @param $queue CRM_Queue_Queue, (for 'enqueue') the modifiable list of pending up upgrade tasks
- *
- * @return mixed  based on op. for 'check', returns array(boolean) (TRUE if upgrades are pending)
- *                for 'enqueue', returns void
+ * Implements hook_civicrm_upgrade().
```
tunbola: Please add the parameter doc block back.

Did you make this changes manually or you used a code sniffer to correct automatically?

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -201,24 +190,20 @@ function tasksassignments_civicrm_pageRun($page) {
...
- * Implementation of hook_civicrm_tabset.
+ * Implements hook_civicrm_tabset().
  *
  * Tasks and documents tabs should appear after Assignments tab directly
  * and since assignments tab weight is set to 30 we set both of those to
  * 40 and 50 respectively.
  *
- * @param string $tabsetName
- * @param array &$tabs
- * @param array $context
```
tunbola: Please add the param doc blocks back.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -245,10 +230,7 @@ function tasksassignments_civicrm_tabset($tabsetName, &$tabs, $context) {
 }
 
 /**
- * Implementation of hook_civicrm_permission
- *
- * @param array $permissions
- * @return void
+ * Implements hook_civicrm_permission().
```
tunbola: Missing parameter doc block.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -299,8 +283,7 @@ function tasksAssignments_civicrm_alterAngular(\Civi\Angular\Manager $angular) {
 function tasksAssignments_civicrm_angularModules(&$angularModules) {
   $angularModules['crm-tasks-workflows'] = array(
```
tunbola: Can we use the short array syntax here.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -524,25 +507,32 @@ function _tasksAssignments_allow_only_add_timeline_action (phpQueryObject $doc)
...
-
-    $columnHeader->remove();
-    $columnBody->remove();
+    _tasksAssignments_remove_column($doc, $columnName);
   }
 }
 
+/**
+ * Removes the category header from the workflow configuration screen.
+ */
```
tunbola: Missing parameter doc block.

igorpavlov: I think params description is missing.

uk.co.compucorp.civicrm.tasksassignments/js/src/crm-tasks-workflows/decorators/route.decorator.js
reneolivo: I have refactored it this way to avoid injection issues:

```js
function getResolverForWorkflowTypes () {
  return ['crmApi', function workflowTypesResolver (crmApi) {
    // ...
  });
}
```

reneolivo:
@tunbola the removal of the parameters was a suggestion by `civilint` since they are supposedly already documented at the original hook definition. Since we don't have an auto-generated documentation it does not make sense to remove these parameters so I restored them back (I amended the commit).

Also, I added a bit of safety in case the table header is not present, like this:

```php
function _tasksAssignments_remove_column(phpQueryObject $doc, $columnName) {
  $columnHeader = $doc->find('table th:contains("' . $columnName . '")');

  // skip if the column does not exist:
  if (!count($columnHeader)) {
    return;
  }

  // do the removal
}
```

Since otherwise it would throw an error when trying to find the index of an element that does not exist.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -39,9 +39,9 @@ function tasksassignments_civicrm_apiWrappers(&$wrappers, $apiRequest) {
 }
 
 /**
- * Implementation of hook_civicrm_xmlMenu
+ * Implements hook_civicrm_xmlMenu().
  *
- * @param $files array(string)
+ * @param array(string) $files
```
tunbola: array(string) is not a valid param type.
Just make it array.

uk.co.compucorp.civicrm.tasksassignments/tasksassignments.php
```diff
@@ -524,25 +525,41 @@ function _tasksAssignments_allow_only_add_timeline_action (phpQueryObject $doc)
...
+ * @param phpQueryObject $doc
+ * @param string $columnName
+ */
+function _tasksAssignments_remove_column(phpQueryObject $doc, $columnName) {
+  $columnHeader = $doc->find('table th:contains("' . $columnName . '")');
 
-    $columnHeader->remove();
-    $columnBody->remove();
+  // skip if the column does not exist:
+  if (!count($columnHeader)) {
```
tunbola: Seems `$columnHeader` is an object and not an array, count is usually associated with an array.
This should be enough:
```php
if(!$columnHeader) {
return;
}
```

PR #357, opened 2018-04-30
PCHR-3540: Delete the open case activity

reneolivo:
## Overview
This PR removes the *Open Case* activities from new and existing Case Type's timelines

### New Case Types
before-new-case-type

### Existing Case Types
before-edit-case-type

### New Case Types
after-new-case-type

### Existing Case Types
after-edit-case-type

### Removing from new Case Types
In the original Case Activity controller, the open case is added from a template constant:
 
```js
var newCaseTypeTemplate = {
  // ...
  definition: {
    // ...
    activitySets: [
      {
        name: 'standard_timeline',
        // ...
        activityTypes: [
          {name: 'Open Case', status: 'Completed' } // constant name
        ]
      }
    ],
    // ...
  }
};

// ...
// when the case type is new it clones the case type template:
$scope.caseType = apiCalls.caseType ? apiCalls.caseType : _.cloneDeep(newCaseTypeTemplate);
```
Source: https://github.com/civicrm/civicrm-core/blob/master/ang/crmCaseType.js#L6-L28

To override this the our Case Type extended controller removes all activity types from the template after the controller is initialized:

```js
(function init () {
  var isNewCaseType;

  initParentController();

  // ...

  if (isNewCaseType) {
    clearAllActivityTypesFromMainActivitySet();
  }
})();

function clearAllActivityTypesFromMainActivitySet () {
  $scope.caseType.definition.activitySets[0].activityTypes = [];
}
```

### Removing the Open Case from existing Case Types

A new upgrader was added:

```php
public function upgrade_1036() {
  $caseTypes = civicrm_api3('CaseType', 'get', ['sequential' => 1]);
  $needsToRemoveOpenType = FALSE;

  foreach ($caseTypes['values'] as $caseType) {
    $activitySets = $caseType['definition']['activitySets'];

    foreach ($activitySets as $setIndex => $set) {
      $openCaseIndex = array_search(self::OPEN_CASE_TYPE, array_column($set['activityTypes'], 'name'));

      if ($openCaseIndex !== FALSE) {
        $needsToRemoveOpenType = TRUE;

        // needs to unset the full path for the activity to be properly removed:
        unset($caseType['definition']['activitySets'][$setIndex]['activityTypes'][$openCaseIndex]);
      }
    }

    // Removes Open Case activities from the Case Type only if they include this activity type:
    if ($needsToRemoveOpenType) {
      civicrm_api3('CaseType', 'create', $caseType);
    }
  }

  return TRUE;
}
```

The Ope Case activity was not removed from the option values because this causes the application to break:

case-type

This is because the Case Type controller expects the 'Open Case' activity to exist when it's being initialized ([see here](https://github.com/civicrm/civicrm-core/blob/be5aae338c4f9e20f5166c10530e9391953f6785/ang/crmCaseType.js#L154)). We can't completely remove the core Case Type controller since we only override some of its functionality.


uk.co.compucorp.civicrm.tasksassignments/js/src/crm-tasks-workflows/controllers/casetype.extended.controller.js
```diff
@@ -26,7 +37,7 @@
      */
     function addActivity (activitySet, activityType) {
       // call parent
```
igorpavlov: Could you remove this comment, it is not needed anymore.

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
```diff
@@ -23,14 +36,43 @@
...
+    });
+
+    describe('when creating a new case type', function () {
+      beforeEach(function () {
+        $scope = $rootScope.$new();
+
+        initController($scope);
+      });
+
+      it('it clears all activities from the main activity set', function () {
```
igorpavlov: I don't think we add "it", so `it('clears all activities from the main activity set'` should be enough. Same for other places.

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
```diff
@@ -98,5 +142,20 @@
 
       return taskOptions.concat(documentOptions);
     }
+
+    /**
+     * Returns a mock Case Types
```
igorpavlov: maybe "Returns a mock for Case Types"

uk.co.compucorp.civicrm.tasksassignments/js/src/crm-tasks-workflows/controllers/casetype.extended.controller.js
reneolivo: thanks! crept from another PR that can't be merged for now

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
reneolivo: you you are right.

PR #359, opened 2018-05-04
PCHR-3418: Modify the "Add Activity" modal

reneolivo:
## Overview
This PR makes the following changes to the "Add Activity" modal on the activity types configuration screen:

* It renames the modal title to "Add New Task / Document Type".
* It renames the "Component" field to "Category".
* It only displays "Tasks" and "Documents" in the "Category" options.
* It moves the "Category" field to the top of the form

## Before
activity-before

## After
activity-after

## Technical Details
The code that makes all the changes reads as follows:
```php
public function buildQuickForm() {
  parent::buildQuickForm();

  // Changes the title of the modal:
  CRM_Utils_System::setTitle('Add New Task / Document Type');

  // Adds the "Tasks" and "Documents" options to the "Category" dropdown:
  $categoryOptions = array();

  $civiTaskId = CRM_Core_Component::getComponentID('CiviTask');
  $civiDocumentId = CRM_Core_Component::getComponentID('CiviDocument');

  if ($civiTaskId !== NULL) {
    $categoryOptions[$civiTaskId] = t('Tasks');
  }

  if ($civiDocumentId !== NULL) {
    $categoryOptions[$civiDocumentId] = t('Documents');
  }

  $this->add('select',
    'component_id',
    ts('Category'),
    $categoryOptions, FALSE
  );

  // ...
}
```

The following code was removed since it was leaving category options for other components that are not relevant to Tasks and Workflows:

```php
// removed this:
$newOptions = array();
if ($this->elementExists('component_id')) {
  $componentSelectElement = $this->getElement('component_id');
  $options = $componentSelectElement->_options;

  foreach ($options as $value) {
    $newOptions[$value['attr']['value']] = $value['text'];
  }
}
```

To change the order of the category field a JS file was injected into the form template:

```php
    Civi::resources()->addScriptFile('uk.co.compucorp.civicrm.tasksassignments', 'js/CRM/Form/Options.js');
```

```js
'use strict';

(function init () {
  var componentIdRow = CRM.$('select#component_id').parents('tr');
  var tableBody = componentIdRow.parents('tbody');

  // Move the category field to the top of the form:
  componentIdRow.detach().prependTo(tableBody);
})();
```

This approach was chosen because the core form template specifies each individual field manually. For example:
https://github.com/civicrm/civicrm-core/blob/3d04355de56e2293afda299bc4e00d5816010082/templates/CRM/Admin/Form/Options.tpl#L58-L63

This makes the whole form hard to extend and using Javascript to move the field is the easiest solution and more future proof.

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Form/Options.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Form/Options.php
@@ -75,6 +75,8 @@ class CRM_Tasksassignments_Form_Options extends CRM_Admin_Form_Options {
       $categoryOptions, FALSE
     );
 
+    Civi::resources()->addScriptFile('uk.co.compucorp.civicrm.tasksassignments', 'js/CRM/Form/Options.js');
+
     //$this->addFormRule(array('CRM_Admin_Form_Options', 'formRule'), $this);
   }
 
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Form/Options.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Form/Options.php
@@ -40,7 +40,7 @@ class CRM_Tasksassignments_Form_Options extends CRM_Admin_Form_Options {
    *
    * @return void
    */
-  function setDefaultValues() {
+  public function setDefaultValues() {
     $defaults = parent::setDefaultValues();
     return $defaults;
   }
@@ -53,33 +53,35 @@ class CRM_Tasksassignments_Form_Options extends CRM_Admin_Form_Options {
    */
   public function buildQuickForm() {
     parent::buildQuickForm();
-    
+
     $newOptions = array();
+
     if ($this->elementExists('component_id')) {
-        $componentSelectElement = $this->getElement('component_id');
-        $options = $componentSelectElement->_options;
-        foreach ($options as $value) {
-          $newOptions[$value['attr']['value']] = $value['text'];
-        }
+      $componentSelectElement = $this->getElement('component_id');
+      $options = $componentSelectElement->_options;
+
+      foreach ($options as $value) {
+        $newOptions[$value['attr']['value']] = $value['text'];
+      }
     }
-    
+
     $civiTaskId = CRM_Core_Component::getComponentID('CiviTask');
-    if ($civiTaskId !== null)
-    {
-        $newOptions[$civiTaskId] = t('Tasks and Assignments');
-    }
     $civiDocumentId = CRM_Core_Component::getComponentID('CiviDocument');
-    if ($civiDocumentId !== null)
-    {
-        $newOptions[$civiDocumentId] = t('Documents and Assignments');
+
+    if ($civiTaskId !== NULL) {
+      $newOptions[$civiTaskId] = t('Tasks and Assignments');
     }
-    
+
+    if ($civiDocumentId !== NULL) {
+      $newOptions[$civiDocumentId] = t('Documents and Assignments');
+    }
+
     $this->add('select',
       'component_id',
       ts('Component'),
       $newOptions, FALSE
     );
-    
+
     //$this->addFormRule(array('CRM_Admin_Form_Options', 'formRule'), $this);
   }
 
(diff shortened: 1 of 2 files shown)


uk.co.compucorp.civicrm.tasksassignments/js/CRM/Form/Options.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+(function init () {
```
igorpavlov: `(function init (CRM) {`

since CRM is a global var

uk.co.compucorp.civicrm.tasksassignments/js/CRM/Form/Options.js
```diff
@@ -0,0 +1,9 @@
...
+  var componentIdRow = CRM.$('select#component_id').parents('tr');
+  var tableBody = componentIdRow.parents('tbody');
+
+  // Move the category field to the top of the form:
+  componentIdRow.detach().prependTo(tableBody);
+})();
```
igorpavlov: `})(CRM);`

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Form/Options.php
```diff
@@ -53,33 +53,30 @@ function setDefaultValues() {
...
-        $options = $componentSelectElement->_options;
-        foreach ($options as $value) {
-          $newOptions[$value['attr']['value']] = $value['text'];
-        }
-    }
-    
+
+    CRM_Utils_System::setTitle('Add New Task / Document Type');
+
+    $categoryOptions = array();
```
tunbola: Please use short array syntax here.

uk.co.compucorp.civicrm.tasksassignments/js/CRM/Form/Options.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+(function init () {
+  var componentIdRow = CRM.$('select#component_id').parents('tr');
+  var tableBody = componentIdRow.parents('tbody');
+
+  // Move the category field to the top of the form:
+  componentIdRow.detach().prependTo(tableBody);
```
igorpavlov: I am wondering if `appendTo` would work here https://stackoverflow.com/questions/1279957/how-to-move-an-element-into-another-element?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa could you try please?

uk.co.compucorp.civicrm.tasksassignments/js/CRM/Form/Options.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+(function init () {
+  var componentIdRow = CRM.$('select#component_id').parents('tr');
```
igorpavlov: You don't need `select` since you have an ID. So you need only `#component_id`

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Form/Options.php
```diff
@@ -53,33 +53,30 @@ function setDefaultValues() {
...
     $this->add('select',
       'component_id',
-      ts('Component'),
-      $newOptions, FALSE
+      ts('Category'),
+      $categoryOptions, FALSE
     );
-    
+
+    Civi::resources()->addScriptFile('uk.co.compucorp.civicrm.tasksassignments', 'js/CRM/Form/Options.js');
```
tunbola: Can we also remove the commented line after this, since its not in use anymore.

uk.co.compucorp.civicrm.tasksassignments/js/CRM/Form/Options.js
igorpavlov: Please try to use `closest`, since `parents` will select ALL parent `tr`s.

uk.co.compucorp.civicrm.tasksassignments/js/CRM/Form/Options.js
```diff
@@ -0,0 +1,9 @@
...
+  var tableBody = componentIdRow.parents('tbody');
```
igorpavlov: Same here with `parents`

reneolivo:
### Fixes after QA

* https://github.com/compucorp/civihr-tasks-assignments/pull/359/commits/eefbfe21755fef9c1319d2f4db191b9ac9b72ad5

### Description

The title of the modal now changes if you are creating a new activity type or editing an existing one.

### Technical details

To determine if the form is for a new option value (activity type) or an existing one the ID of the record was queried:

```php
$isNewOptionValue = empty($this->get('id'));

if ($isNewOptionValue) {
  CRM_Utils_System::setTitle('Add New Task / Document Type');
}
else {
  CRM_Utils_System::setTitle('Edit Task / Document Type');
}
```

PR #361, opened 2018-05-14
PCHR-3659: Hide Application case type

deb1990:
## Overview
As part of this PR the "Application" case type has been hidden from the Workflow Types page(civicrm/a/#/caseType).

## Before
2018-05-14 at 4 58 pm

## After
2018-05-14 at 4 57 pm

## Technical Details
As part of T&W Epic the Application case type was supposed to be hidden, but the requirement was missed, hence when clicked on Application, there were console errors.

In PCHR-3369 we have filtered the activity types to `CiviTask` and `CiviDocument` category only. But the existing Application category is saved with values other than the above categories i.e. Phone Call, Email, Meeting and Follow up are not from CiviTask and CiviDocument category hence the error, which is expected.

The Case type has been hidden as per the requirements.


igorpavlov:
Could you please rephrase the PR Overview to smth like:

> This PR hides the Application case type in the Workflow Types form.

Otherwise it sounds like this is not the main change of the PR.

PR #362, opened 2018-05-18
PCHR-3700: Add Activity Type menu item

reneolivo:
## Overview
This PR adds a new menu item for Activity Types under the tasks administration menu. 

## Before
before-menu-item

## After
after-menu-item

## Technical Details

A new upgrader was added:

```php
public function upgrade_1037() {
  civicrm_api3('Navigation', 'create', [
    'label' => 'Activity Types',
    'name' => 'activity_types_administer',
    'url' => 'civicrm/admin/options/activity_type?reset=1',
    'permission' => 'administer CiviCase',
    'parent_id' => 'tasksassignments_administer',
    'is_active' => 1,
  ]);

  return TRUE;
}
```

## Comments
This task is about adding a new menu item, not replacing/moving the old one under `Configure > Customize Data and Screens > Activity Type`

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -895,6 +895,24 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base {
     return TRUE;
   }
 
+  /**
+   * Adds a new menu item for Activity Types under the tasks administration menu.
+   *
+   * @return bool
+   */
+  public function upgrade_1037() {
+    civicrm_api3('Navigation', 'create', [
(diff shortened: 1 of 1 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -908,7 +908,7 @@ public function upgrade_1037() {
     // Create the menu item only if it doesn't exist:
     if ($activityTypeMenuItem['count'] === 0) {
       civicrm_api3('Navigation', 'create', [
-        'label' => 'Activity Types',
+        'label' => 'Task and Document Types',
         'name' => 'activity_types_administer',
```
tunbola: Shouldn't we change the name too to reflect the label. Since the former name reflects the label.

PR #363, opened 2018-05-25
PCHR-3714: Task Notification Badge Component

deb1990:
## Overview
As part of this PR a notification Badge has been added to the `Tasks` menu of SSP.
**Related PRs**
https://github.com/civicrm/civihr/pull/2623
https://github.com/compucorp/civihr-employee-portal/pull/469
https://github.com/compucorp/civihr-employee-portal-theme/pull/348

## Screenshot
after

## Technical Details
1. A new module `tasks-notificaction-badge` and a component with same name, has been created. Also the unit test has been setup.

## Comments
In https://github.com/compucorp/civihr-tasks-assignments/commit/d077ee62e42902c241992b7b28795099057c8d9a#diff-b63b868cf433e7970d2aa11e0cd17e7bR167 `$params['check_permissions']` was changed to `TRUE` by mistake, which has been reverted here.


uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-notification-badge/components/tasks-notification-badge.component.js
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Get the logged in contact id and save it as assignee
+     *
+     * @returns {Promise}
+     */
+    function getAssigneeUserId () {
+      return Session.get()
+        .then(function (session) {
+          taskFilters.params.assignee_contact_id = session.contactId;
```
reneolivo: are you really getting the user session id or are you instead appending the user id to the filters?

I suggest the init be refactored like this instead:

```js
(function init () {
  getAssigneeUserId()
    .then(function (assigneeContactId) {
      taskFilters.params.assignee_contact_id = assigneeContactId;
      documentFilters.params.assignee_contact_id = assigneeContactId;
      vm.filters = [taskFilters, documentFilters];
    });
})();

function getAssigneeUserId () {
  return Session.get()
    .then(function (session) {
      return session.contactId;
    });
}
```

uk.co.compucorp.civicrm.tasksassignments/js/test/tasks-notification-badge/component/tasks-notification-badge.component.spec.js
```diff
@@ -0,0 +1,52 @@
...
+    it('is initialized', function () {
+      expect($log.debug).toHaveBeenCalled();
+    });
+
+    it('filters the Tasks where status is not completed and assignee is the logged in user', function () {
+      expect(controller.filters[0]).toEqual({
+        apiName: 'Task',
+        params: {
+          status_id: { '!=': 'Completed' },
+          assignee_contact_id: window.Drupal.settings.currentCiviCRMUserId
```
reneolivo: Although technically correct, you better include the Session service and mock the response, since this is what you are ultimately using inside the badge implementation. If the session service happens to change and uses another method to determine the session it would break this test.

uk.co.compucorp.civicrm.tasksassignments/css/civitasks.css
```diff
@@ -12853,15 +12853,17 @@ fieldset[disabled]
   border: 0;
   padding: 10px !important;
 }
-.mobile [type='date'][uib-datepicker-popup] {
+#civitasks .___outside-namespace .mobile [type='date'][uib-datepicker-popup], #cividocuments .___outside-namespace .mobile [type='date'][uib-datepicker-popup] {
   line-height: normal;
+  /* stylelint-disable selector-max-compound-selectors */
+  /* stylelint-enable */
```
reneolivo: what caused this file to change?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-notification-badge.js
```diff
@@ -0,0 +1,19 @@
...
+
+  require([
+    'tasks-notification-badge/app'
+  ],
+  function (angular) {
+    angular.bootstrap(
+      document.querySelector('[data-tasks-notification-badge]'), ['tasks-notification-badge']
+    );
+  });
+})(CRM, require);
```
reneolivo: why are we creating a new module just for these notifications? Why not just add the notifications inside `/js/src/tasks-assignments/components` ?

reneolivo:
Side question @deb1990 : what happens when a workflow is created? That uses a different Case endpoint, but when creating a case a bunch of tasks are created too. Has that been taken into account?

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-notification-badge.js
deb1990: We need a new module, because in the end we need to create a separate `.min.js` which would be imported by the SSP Menu. Reference https://github.com/compucorp/civihr-employee-portal/pull/469/files#diff-0629dcabaa6a0ce1bb40f802dbd6e24cR6574

If we dont do this, we would need to import the whole T&A module just for the small badge. 

The same was followed in L&A too, where we created a new module named `manager-notification-badge`.

uk.co.compucorp.civicrm.tasksassignments/js/src/tasks-notification-badge/components/tasks-notification-badge.component.js
deb1990: I decided to use `Session.get` directly, without a separate function.

deb1990:
@reneolivo 

> Side question @deb1990 : what happens when a workflow is created? That uses a different Case endpoint, but when creating a case a bunch of tasks are created too. Has that been taken into account?

Yes it works too, I tested it. Because in the end the component fetches information from the Tasks API, so no matter how the tasks are added, the component is going to work as far as the API returns correct values.

uk.co.compucorp.civicrm.tasksassignments/css/civitasks.css
deb1990: Really thanks for noticing this bug. I have created a ticket PCHR-3724, to fix this across all repos. Please check the ticket for more details.

PR #366, opened 2018-05-30
PCHR-3732: Fetch only current version of Documents

deb1990:
## Overview
When Documents are added via Add Workflow, and then documents are uploaded from SSP, the Tasks Notification Badge count does not get updated.

## Before
b

## After
a

## Technical Details
1. When documents are uploaded from SSP, a new revision of the Document is created in the backend, but still the old record stays in the DB. Hence it is required to fetch only the current revisions.

Added the same parameter in `tasks-notification-badge.component.js`


uk.co.compucorp.civicrm.tasksassignments/js/test/tasks-notification-badge/component/tasks-notification-badge.component.spec.js
```diff
@@ -52,7 +52,8 @@ define([
           apiName: 'Document',
           params: {
             status_id: 'awaiting upload',
-            assignee_contact_id: loggedInUserID
+            assignee_contact_id: loggedInUserID,
+            is_current_revision: true
```
AkA84: you need to update the `it` description

deb1990:
> The spinner needs to be reduced to a size that doesn't scare the user
The spinner should keep on spinning until the modal closes, while right now it disappears earlier

Fixed in https://github.com/compucorp/civihr-employee-portal-theme/pull/352

PR #374, opened 2018-06-13
PCHR-3836: Add Upgrader for Case Type Custom Field

ajesamson:
## Overview
For proper grouping of case type into categories, custom field need to be setup. This PR creates upgrader for setting up custom field with option values  `Workflow and Vacancy`.

## Before
There was no custom group and field relating to case types.

## After
New custom group (`Case Type Category`) and custom field (`Category`) was setup with options workflow and vacancy for categorization. 

## Technical Details
Creating the custom fields automatically setup the two options (`workflow and vacancy`) for since the the `html_type` for the custom field is `Select`.

```php
private function _createCustomField() {
  civicrm_api3('CustomField', 'create', [
    'custom_group_id' => 'case_type_category',
    'label' => 'Category',
    'name' => 'category',
    'data_type' => 'String',
    'html_type' => 'Select',
    'is_required' => 1,
    'column_name' => 'category',
    'option_values' => array('Workflow' => 'Workflow', 'Vacancy' => 'Vacancy'),
  ]);
}
```


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -919,13 +919,86 @@ public function upgrade_1037() {
...
     CRM_Core_DAO::executeQuery("DELETE FROM `civicrm_navigation` WHERE name IN ('tasksassignments', 'ta_dashboard_tasks', 'ta_dashboard_documents', 'ta_dashboard_calendar', 'ta_dashboard_keydates', 'tasksassignments_administer', 'ta_settings')");
     CRM_Core_BAO_Navigation::resetNavigation();
 
     return TRUE;
   }
+  
+  /**
+   * Creates case type option value for cg_extend_objects option group
+   */
+  private function _createOptionValue() {
```
davialexandre: The private modifier already takes care of making this private, so there's no need to start the method names with an underscore

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -919,13 +919,86 @@ public function upgrade_1037() {
...
+  private function _createCustomField() {
+    civicrm_api3('CustomField', 'create', [
+      'custom_group_id' => 'case_type_category',
+      'label' => 'Category',
+      'name' => 'category',
+      'data_type' => 'String',
+      'html_type' => 'Select',
+      'is_required' => 1,
+      'column_name' => 'category',
+      'option_values' => array('Workflow' => 'Workflow', 'Vacancy' => 'Vacancy'),
```
davialexandre: Instead of hardcoding these options here, would it be possible to just point this to an option group where we can add new option values in the future?

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
ajesamson: Great. Will update

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
ajesamson: I tested again and we can replace the `option_values` with `option_group_id` of an existing record. The ticket requirement was to create these two values. An option could have been to add another implementation for setting up separate option group and values or allow the custom field have setup to auto-generate the group and values at go.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -919,13 +919,140 @@ public function upgrade_1037() {
...
+    ]);
+    if ($customGroups['count'] == 0) {
+      $this->createCustomGroup();
+    }
+    
+    $customFields = civicrm_api3('CustomField', 'get', [
+      'custom_group_id' => 'case_type_category',
+    ]);
+    if ($customFields['count'] == 0) {
+      $optionGroupId = $this->createCustomFieldOptionValues();
```
davialexandre: Same thing here. Perhaps it would be better to rename this to something like `createCaseTypeCategoryOptionValues()`

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -919,13 +919,140 @@ public function upgrade_1037() {
...
+      if ($optionGroupId == null) {
+        return FALSE;
+      }
 
+      $this->createCustomField($optionGroupId);
```
davialexandre: And here too. Perhaps `createCaseTypeCategoryCustomField()`

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -919,13 +919,140 @@ public function upgrade_1037() {
...
+  
   public function uninstall() {
     CRM_Core_DAO::executeQuery("DELETE FROM `civicrm_navigation` WHERE name IN ('tasksassignments', 'ta_dashboard_tasks', 'ta_dashboard_documents', 'ta_dashboard_calendar', 'ta_dashboard_keydates', 'tasksassignments_administer', 'ta_settings')");
     CRM_Core_BAO_Navigation::resetNavigation();
 
     return TRUE;
   }
+  
+  /**
+   * Sets up option group and values used for case type custom field
```
davialexandre: s/case type custom field/case type category custom field/

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -919,13 +919,140 @@ public function upgrade_1037() {
...
+   * @return null
+   */
+  private function createCustomFieldOptionValues() {
+    $result = civicrm_api3('OptionGroup', 'get', [
+      'name' => 'case_type_category',
+    ]);
+    if ($result['count'] == 0) {
+      $optionValues = ['Workflow', 'Vacancy'];
+      $groupParams = [
+        'name' => 'category_custom_field',
```
davialexandre: I'm confused. In line 981 you try to fetch an Option group named `case_type_category`, but here, if that doesn't exist, you try to create a group named `category_custom_field`? Is this right? Shouldn't both be the same?

PR #376, opened 2018-06-18
PCHR-3543: Update T&A Settings

backstageel:
## Overview
Set default number of days before document expiry when the clone document should be created = 90 days
If this field is blank, update it to 90 days.

## Before
There was an Upgrader function "upgrade_1020" that was setting default days before document expiry to 0.

## After
The new Upgrader function "upgrade_1038" is setting the default number of days to 90 if it is empty or equals to zero. And the existing upgrader  "upgrade_1020" was changed to set default number of days to 90.

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -928,10 +928,13 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base {
    */
   public function upgrade_1038() {
     $optionValue = civicrm_api3('OptionValue', 'get', [
-      'sequential' => 1,
       'option_group_id' => 'ta_settings',
       'name' => 'days_to_create_a_document_clone',
     ]);
+    if(empty($optionValue['id'])) {
+      return TRUE;
+    }
+
     $optionValue = array_shift($optionValue['values']);
     if (empty($optionValue['value'])) {
       civicrm_api3('OptionValue', 'create', [
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -920,6 +920,29 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base {
     return TRUE;
   }
 
+  /**
+   * Set default number of days before document expiry when the clone document
+   * should be created = 90 days If this field is blank, update it to 90 days.
+   *
+   * @return bool
+   */
+  public function upgrade_1038() {
+    $optionValue = civicrm_api3('OptionValue', 'get', [
+      'sequential' => 1,
+      'option_group_id' => 'ta_settings',
+      'name' => 'days_to_create_a_document_clone',
+    ]);
+    $optionValue = array_shift($optionValue['values']);
+    if (empty($optionValue['value'])) {
+      civicrm_api3('OptionValue', 'create', [
+        'id' => $optionValue['id'],
+        'value' => 90,
+      ]);
+    }
+
+    return TRUE;
+  }
+
   public function uninstall() {
     CRM_Core_DAO::executeQuery("DELETE FROM `civicrm_navigation` WHERE name IN ('tasksassignments', 'ta_dashboard_tasks', 'ta_dashboard_documents', 'ta_dashboard_calendar', 'ta_dashboard_keydates', 'tasksassignments_administer', 'ta_settings')");
     CRM_Core_BAO_Navigation::resetNavigation();


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -920,6 +920,39 @@ public function upgrade_1037() {
...
+   * Set default number of days before document expiry when the clone document
+   * should be created = 90 days If this field is blank, update it to 90 days.
+   *
+   * @return bool
+   */
+  public function upgrade_1038() {
+    $optionValue = civicrm_api3('OptionValue', 'get', [
+      'sequential' => 1,
+      'option_group_id' => 'ta_settings',
+      'name' => "days_to_create_a_document_clone",
```
tunbola: Please use single quotes here.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -920,6 +920,39 @@ public function upgrade_1037() {
...
+    ]);
+    if (empty($optionValue['id'])) {
+      civicrm_api3('OptionValue', 'create', [
+        'option_group_id' => 'ta_settings',
+        'name' => 'days_to_create_a_document_clone',
+        'label' => is_callable('t') ? t('Renewed document creation date offset (days)') : 'Renewed document creation date offset (days)',
```
tunbola: Any reason for not using the translation function without checking if its callable?
I think there's no need for this, just use it. Its a civi function and should always be callable.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -920,6 +920,39 @@ public function upgrade_1037() {
...
+   *
+   * @return bool
+   */
+  public function upgrade_1038() {
+    $optionValue = civicrm_api3('OptionValue', 'get', [
+      'sequential' => 1,
+      'option_group_id' => 'ta_settings',
+      'name' => "days_to_create_a_document_clone",
+    ]);
+    if (empty($optionValue['id'])) {
```
tunbola: do we need this logic to create the option value at all? The upgrader: `upgrade_1020` would have been ran for all sites and this option value would have been created for all sites.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -512,7 +512,7 @@ public function upgrade_1020() {
         'option_group_id' => 'ta_settings',
         'name' => 'days_to_create_a_document_clone',
         'label' => is_callable('t') ? t('Renewed document creation date offset (days)') : 'Renewed document creation date offset (days)',
-        'value' => 0,
+        'value' => 90,
```
tunbola: We shouldn't modify an upgrader that has already been ran on sites.. Rather we should create a new one to effect any change we need.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -920,6 +920,39 @@ public function upgrade_1037() {
...
+      civicrm_api3('OptionValue', 'create', [
+        'option_group_id' => 'ta_settings',
+        'name' => 'days_to_create_a_document_clone',
+        'label' => is_callable('t') ? t('Renewed document creation date offset (days)') : 'Renewed document creation date offset (days)',
+        'value' => 90,
+      ]);
+    }
+    else {
+      $optionValue = array_shift($optionValue['values']);
+      if (empty($optionValue['value']) || $optionValue['value'] == 0) {
```
tunbola: Checking if its empty should be enough, it will also cater for the situation where `$optionValue['value'] == 0`
http://php.net/manual/en/function.empty.php

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
backstageel: I copied from another upgrader and wondered if i should change it... But will change now

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -920,6 +920,29 @@ public function upgrade_1037() {
...
 
+  /**
+   * Set default number of days before document expiry when the clone document
+   * should be created = 90 days If this field is blank, update it to 90 days.
+   *
+   * @return bool
+   */
+  public function upgrade_1038() {
+    $optionValue = civicrm_api3('OptionValue', 'get', [
+      'sequential' => 1,
```
tunbola: I don't think we need sequential here.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -920,6 +920,29 @@ public function upgrade_1037() {
...
+      'option_group_id' => 'ta_settings',
+      'name' => 'days_to_create_a_document_clone',
+    ]);
```
tunbola: Would be good to check 
```php
if(empty($optionValue['id'])) {
return TRUE;
}
```
I know it should exist but just good to write the code in a safe way.

tunbola:
Typo in the PR description s/or iquals to zero/or equals to zero

PR #380, opened 2018-06-26
PCHR-3429: CiviCRM upgrade

davialexandre:
## Overview

This PR is part of https://github.com/compucorp/civihr/pull/2720 and updates Tasks and Assignments so that it can be used with CiviCRM 5.3.0. It includes fixes for things that stopped working after the upgrade, syncing of overridden files and some minor fixes not related to the upgrade. See "Technical Details" for more information.

## Before

The extension could not be used with CiviCRM 5.3.0

## After

The extension can be used with CiviCRM 5.3.0

### Refactoring of the TASettings API

This extension has its own API to save and fetch settings: `TASettings`. Internally, this API would use Option Values to store the settings. That is, each setting was an Option Value in the `ta_settings` Option Value, and the value was the setting value. Whenever a setting was updated, the Option Value value was updated as well. Because of that, it was possible to have more than one Option Value with the same value and since it's not possible anymore to have that (https://github.com/civicrm/civicrm-core/pull/11089), the installation was failing.

The reasons for this implementation are unknown, as the code is quite old and undocumented. Maybe it made sense back when it was implemented, but today the proper solution for this is to use [CiviCRM settings](https://docs.civicrm.org/dev/en/latest/framework/setting/) and the `TASettings` API was refactored to use that:

- First, the code style was fixed to use the conventions we use today https://github.com/compucorp/civihr-tasks-assignments/pull/380/commits/6fbbb190bd447cb1b17e2ca43417220ebd90d6ab
- Next, tests were added to cover the existing behavior and make sure not would break after the refactoring: https://github.com/compucorp/civihr-tasks-assignments/pull/380/commits/ce858eca5c0bb1698a814a84c61db08683e3ca2d
- Then a new `ta_settings` setting was added to store the values using CiviCRM Settings. The internal implementation of the API was updated to use that instead of Option Values: https://github.com/compucorp/civihr-tasks-assignments/pull/380/commits/ca56053096155e0d88741560177438ad55e0e1c2
- Next, I made sure the `ta_settings` Option Group would not be created on new installs: https://github.com/compucorp/civihr-tasks-assignments/pull/380/commits/bd865064149b95382adb16e1e52dddc9e68d5c26
- And finally, for existing sites, migrated the settings in Option Values to CiviCRM settings and deleted the Option Group.

While doing that, I also made one small improvement to the API where I validated the settings before saving them. This was necessary because the `ta_settings` setting is an array and without the validation, any field passed to it would end up being stored in the settings. The validation was added here: https://github.com/compucorp/civihr-tasks-assignments/pull/380/commits/b6e6bc5a0524aad30cead2e71c957b0b26c7d1fd

### Syncing overridden files

The extension has only one overridden file to be synced:

- `CRM/Activity/Form/ActivityView.php`: [diff](https://www.diffchecker.com/S9UswuPk)

### Default date in the Assignment Fabricator

This is the code that would set the default `start_date` for an Assigment:

```
'start_date' => ('Y-m-d H:i:s')
```

The problem is that `('Y-m-d H:i:s')` is not a valid date and when passed to the API to build the SQL query, that would be converted to 0 which would result in a database error. 

To fix that, I simply changed the line to:

```
'start_date' => date('Y-m-d H:i:s')
```

I've spend quite a lot of time trying to figure out why this never failed before, but I couldn't. The most likely reason is that something about saving cases changed in civicrm-core but I'm not sure.

### DIFF
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -853,6 +853,41 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base {
     return TRUE;
   }
 
+  /**
+   * Migrate the T&A Settings from Option Values to CiviCRM Settings.
+   *
+   * As part of the migration, the Option Values (and their Option Group)
+   * are deleted.
+   *
+   * @return bool
+   * @throws \CiviCRM_API3_Exception
+   */
+  public function upgrade_1039() {
+    // We need to flush the cache to make sure the settings
+    // metadata in settings/taskandassignments.settings.php
+    // will be read by Civi
+    CRM_Utils_System::flushCache();
+
+    $taSettingsOptionValues = $this->_getTaSettingsOptionValues();
+
+    if ($taSettingsOptionValues === NULL) {
+      return TRUE;
+    }
+
+    $settings = [];
+    foreach ($taSettingsOptionValues as $name => $value) {
+      $settings[$name] = $value;
+    }
+
+    civicrm_api3('TASettings', 'set', [
+      'fields' => $settings
+    ]);
+
+    $this->_deleteTASettingsOptionGroup();
+
+    return TRUE;
+  }
+
   public function uninstall() {
     CRM_Core_DAO::executeQuery("DELETE FROM `civicrm_navigation` WHERE name IN ('tasksassignments', 'ta_dashboard_tasks', 'ta_dashboard_documents', 'ta_dashboard_calendar', 'ta_dashboard_keydates', 'tasksassignments_administer', 'ta_settings')");
     CRM_Core_BAO_Navigation::resetNavigation();
@@ -989,6 +1024,43 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base {
         'label' => $toLabel
       ]);
     }
+  }
+
+  /**
+   * Returns a list of all the Option Values of the ta_settings Option Group.
+   *
+   * The format of the list will be: [ option_value_name => option_value_value ]
+   *
+   * @return array|null
+   *   An array with the Option Values, or NULL in case neither
+   *   the option group or values exist anymore
+   */
+  private function _getTaSettingsOptionValues() {
+    try {
+      $result = civicrm_api3('OptionValue', 'get', [
+        'sequential' => 1,
+        'option_group_id' => 'ta_settings'
+      ]);
+    } catch (CiviCRM_API3_Exception $e) {
+      return NULL;
+    }
+
+    return array_column($result['values'], 'value', 'name');
+  }
+
+  /**
+   * Deletes the ta_settings Option Group
+   *
+   * @throws \CiviCRM_API3_Exception
+   */
+  private function _deleteTASettingsOptionGroup() {
+    civicrm_api3('OptionGroup', 'get', [
+      'sequential' => 1,
+      'name' => 'ta_settings',
+      'api.OptionGroup.delete' => [
+        'id' => '$value.id'
+      ]
+    ]);
   }
 
 }
--- a/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
+++ b/uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
@@ -214,52 +214,8 @@ class CRM_Tasksassignments_Upgrader extends CRM_Tasksassignments_Upgrader_Base {
    * @return bool
    */
   public function upgrade_0007() {
-    $optionGroupID = CRM_Core_DAO::getFieldValue('CRM_Core_DAO_OptionGroup', 'ta_settings', 'id', 'name');
-
-    if (!$optionGroupID) {
-      $params = array(
-        'name' => 'ta_settings',
-        'title' => 'Tasks and Assignments settings',
-        'is_active' => 1,
-        'is_reserved' => 1,
-      );
-
-      civicrm_api3('OptionGroup', 'create', $params);
-
-      $optionsValue = array(
-        'documents_tab' => array(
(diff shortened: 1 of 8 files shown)


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -214,52 +214,8 @@ public function upgrade_0006() {
...
-          'option_group_id' => 'ta_settings',
-          'name' => $key,
-          'label' => $value['label'],
-          'value' => $value['value'],
-        );
-
-        civicrm_api3('OptionValue', 'create', $opValueParams);
-      }
-    }
+    // NOOP
```
mickadoo: The reason for removing the code is clear, I'm curious in cases like this if we should just remove the upgrade function completely. It should be clear from the gap in numbers that something was removed. 

On one hand we have something similar to commented code (empty functions), on the other hand we keep sequential numbers for upgrader functions and have a comment to explain the reason.

I don't mind either way we do this, just wondering what your opinion on it is.

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -927,22 +847,44 @@ public function upgrade_1037() {
...
+  }
+
+  /**
+   * Migrate the T&A Settings from Option Values to CiviCRM Settings.
+   *
+   * As part of the migration, the Option Values (and their Option Group)
+   * are deleted.
+   *
+   * @return bool
+   * @throws \CiviCRM_API3_Exception
```
mickadoo: This came up before, but are we encouraging the use of `@throws`?

uk.co.compucorp.civicrm.tasksassignments/api/v3/TASettings.php
```diff
@@ -1,70 +1,161 @@
...
-            //'label' => $item['label'],
-            'value' => $item['value'] != "null" ? $item['value'] : ""
-        );
+  $fields = _civicrm_api3_t_a_settings_valid_fields();
+
+  if (!empty($params['fields'])) {
+    $fields = (array) $params['fields'];
+  }
+
+  $settings = _civicrm_api3_t_a_settings_get_settings();
```
mickadoo: You only use ` $settings['ta_settings']` after this, so how about shortening the future variables and making `$settings = _civicrm_api3_t_a_settings_get_settings()['ta_settings'];`?

uk.co.compucorp.civicrm.tasksassignments/api/v3/TASettings.php
```diff
@@ -1,70 +1,161 @@
...
-        ));
+  $validFields = _civicrm_api3_t_a_settings_valid_fields();
+
+  $requestFields = [];
+  if (!empty($params['fields']) && is_array($params['fields'])) {
+    $requestFields = $params['fields'];
+  }
+
+  foreach ($requestFields as $field => $value) {
+    if (!in_array($field, $validFields)) {
```
mickadoo: Do you think it might be better to do the validation first before the update? In this case it would be possible to pass an array of multiple valid settings and all those up to the first invalid one would be updated

uk.co.compucorp.civicrm.tasksassignments/api/v3/TASettings.php
```diff
@@ -1,70 +1,161 @@
...
+}
+
+/**
+ * Returns the T&A Settings
+ *
+ * @return array
+ * @throws \CiviCRM_API3_Exception
+ */
+function _civicrm_api3_t_a_settings_get_settings() {
+  $settings = array_shift(civicrm_api3('Setting', 'get', [
```
mickadoo: Could just do a `return` from here

uk.co.compucorp.civicrm.tasksassignments/settings/tasksandassignments.setting.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+  return [
```
mickadoo: I had no idea you could do this, even after reading the [documentation](https://docs.civicrm.org/dev/en/latest/framework/setting/#creating-a-new-setting-in-an-extension) for this I still wasn't clear that this was how you do this :man_shrugging: 

Good to know though

uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
davialexandre: Personally, it's not a big deal for me if we encourage it or not, but I tend to think that we should use it. We haven't agreed on this though, so, for now, I think it's ok if people add it or not

uk.co.compucorp.civicrm.tasksassignments/api/v3/TASettings.php
davialexandre: I optimized this whole thing a little bit. Now, instead of making multiple calls to `Setting.create` it does that only once. It now works as you've suggested and the settings will not be updated if there's something wrong.

uk.co.compucorp.civicrm.tasksassignments/api/v3/TASettings.php
davialexandre: I've updated this function a little bit, so I think that now this is not relevant anymore

uk.co.compucorp.civicrm.tasksassignments/api/v3/TASettings.php
```diff
@@ -132,15 +131,18 @@ function civicrm_api3_t_a_settings_set($params) {
...
  *
  * @return array
- * @throws \CiviCRM_API3_Exception
  */
 function _civicrm_api3_t_a_settings_get_settings() {
-  $settings = array_shift(civicrm_api3('Setting', 'get', [
-    'sequential' => 1,
-    'return' => ['ta_settings']
-  ])['values']);
+  try {
```
mickadoo: Just curious why you decided to put this in a `try/catch` block now

PR #403, opened 2018-07-31
PCHR-4049: Upgrade gulp

igorpavlov:
## Overview

This PR is upgrading Gulp to the version `4.0`.

### Basic changes

Please refer to https://github.com/compucorp/civihr/pull/2804 to see the basic technical details that were also applied in this PR.

### Compass

If Compass or/and Compass-Bootstrap is not yet installed on the computer gulp will not show any error and the css file will not be updated. To prevent this an error is now shown:

```js
// The app style relies on compass's gems,
// so we need to rely on it for the time being
exec('compass compile', function (error, stdout) {
  if (error) {
    console.log(error);
```
----

✅*Gulp* tested (deleted dist files and successfully rebuilt without changes)
✅*Requirejs* tested
✅*Test* tested
✅*Sass* tested
✅*Watch* tested


package.json
```diff
@@ -70,5 +61,16 @@
     "stylelint-order": "^0.8.1",
     "stylelint-scss": "^2.5.0",
     "yargs": "^11.1.0"
+  },
+  "dependencies": {
```
AkA84: these are still dev dependencies, not dependencies

igorpavlov: Nice catch, my bad, I seem to have installed without `--save-dev` flag.

PR #404, opened 2018-08-07
PCHR-4050: Fixes for default assignees

reneolivo:
## Overview
This PR fixes the following issues with Tasks and Documents's assignees:

* it fixes an issue where the name of assignees selected by default would not be displayed.
* it fixes an issue that prevented assignees from being saved to the API.

### Default assignee name
pr-before

### Saving assignees
pr-before-2

### Default assignee name
pr-before-1

### Saving assignees
pr-after-2

### Default assignee name
The name of the default assignee was not being displayed because of a bug on the Contact's API. The bug overwrites the ID of the contact with the ID for their job role if they have any. For example, if the contact's id is `1`, and their job role id is `2`, the API response would return `2` instead of `1` for the id field of the contact. This issue has been reported and will be worked in a different ticket.

To fix our particular case, the API also returns a `contact_id` field, which always returns the proper id for the contact:

```js
    return contactService.get({ IN: contactIds })
        .then(function (contacts) {
          var cachedContacts = _.map(contacts, function (contact) {
            return { id: contact.contact_id, label: contact.display_name }; // <--- here we use `contact_id` instead of `id`.
          });

          // ...
        });
```

### Saving assignees

The modal assignment controller was not implementing the right field as specified by the Task API. For sending the data, the specified field is `assignee_id`, as opposed to `assignee_contact_id `:

```js
        // prepares the data before sending:
        var documentListAssignment = vm.documentList.filter(function (doc) {
          return doc.create;
        }).map(function (doc) {
          doc.assignee_id = _.first(doc.assignee_contact_id); // fixed here
          doc.case_id = resultAssignment.id;

          return doc;
        });

        var taskListAssignment = vm.taskList.filter(function (task) {
          return task.create;
        }).map(function (task) {
          task.assignee_id = _.first(task.assignee_contact_id); // fixed here
          task.case_id = resultAssignment.id;

          return task;
        });

        // The data is sent here
```

The tests were also amended. Before they were just making sure that the data to be sent to the API was a clone, but now it checks that the parameters sent are the right ones.


uk.co.compucorp.civicrm.tasksassignments/js/test/tasks-assignments/controllers/modal/modal-assignment.controller.spec.js
```diff
@@ -71,38 +71,60 @@ define([
...
 
-      it('does not pass the original documents objects to the documentService', function () {
-        var documents = documentService.saveMultiple.calls.mostRecent().args[0];
+        expectedDocumentsParameters = getActivitiesToBeSentToAPI(scope.documentList);
+        expectedTasksParameters = getActivitiesToBeSentToAPI(scope.taskList);
 
-        expect(documents.every(function (doc, index) {
-          return doc !== scope.taskList[index];
-        })).toBe(true);
+        scope.confirm() && scope.$digest();
```
igorpavlov: To me this looks like `scope.confirm()` can return `false` and in that case no digest needed. I think I can see the reason you did that - to save 1 line of code. But since we use such syntax for conditions - better to use 2 lines to avoid confusion.

uk.co.compucorp.civicrm.tasksassignments/js/test/tasks-assignments/controllers/modal/modal-assignment.controller.spec.js
```diff
@@ -71,38 +71,60 @@ define([
...
-        })).toBe(true);
+      it('saves every task', function () {
+        expect(taskService.saveMultiple).toHaveBeenCalledWith(expectedTasksParameters);
       });
 
+      /**
+       * Returns a list of activities containing the fields required by their
+       * respective services. It will only return activities marked for creation.
+       *
+       * @param {Array} activities a list of activities that are missing their assignee and case ids.
```
igorpavlov: Please add another space after `@param` to aligh with `@return`

igorpavlov: `* @param  {Array} activities a list of activities missing their assignee and case IDs.`

just a bit more succinct version.

uk.co.compucorp.civicrm.tasksassignments/js/test/tasks-assignments/controllers/modal/modal-assignment.controller.spec.js
```diff
@@ -71,38 +71,60 @@ define([
...
+       * @return {Array}
+       */
+      function getActivitiesToBeSentToAPI (activities) {
+        return _.chain(activities).cloneDeep()
```
igorpavlov: could you push `.cloneDeep` to a separate line please? That would show the chain better.

uk.co.compucorp.civicrm.tasksassignments/js/test/tasks-assignments/controllers/modal/modal-assignment.controller.spec.js
```diff
@@ -114,7 +136,10 @@ define([
         });
 
         scope.taskList.forEach(function (task) {
-          _.assign(task, { activity_date_time: '2013-01-01', assignee_contact_id: [3] });
+          _.assign(task, {
+            assignee_contact_id: [ expectedContactId ],
+            activity_date_time: '2013-01-01'
```
igorpavlov: Could we please move '2013-01-01' to a variable somewhere else so we don't have that constant. Something like:

`var sampleDate = '2013-01-01';`

In that case I understand this is just a sample date and it can be essentially any date and there is nothing special about 2013 New Year.

uk.co.compucorp.civicrm.tasksassignments/js/test/tasks-assignments/controllers/modal/modal-assignment.controller.spec.js
```diff
@@ -410,7 +435,11 @@ define([
       });
 
       describe('after selecting a default assignee', function () {
-        var contact = { id: _.uniqueId(), display_name: 'Jon Snow' };
+        var contact = {
+          contact_id: _.uniqueId(),
+          display_name: 'Jon Snow',
```
igorpavlov: I personally prefer Aria, but OK, let it be John. But could we also move this to a var somewhere in the top of the file? In addition to a sample date we may even have an object:

```js
var sampleData = {
  date: '2013-01-01',
  contactName: 'John Snow'
};
```

uk.co.compucorp.civicrm.tasksassignments/js/test/tasks-assignments/controllers/modal/modal-assignment.controller.spec.js
reneolivo: Thanks, it was like that, I should have fixed it. I removed the `scope.$digest()` statement because it's not needed.

igorpavlov:
Approved codewise, but please check why PHP tests fail (I know there are no PHP changes in this PR, but we cannot send failing code to QA)

https://jenkins.compucorp.co.uk/job/Tasks%20and%20Assignments/job/PR-404/2/display/redirect

PR #408, opened 2018-08-12
PCHR-4111: Hide relationship types that are disabled

reneolivo:
## Overview
This PR hides relationship types that have been disabled. The relationship types are displayed when selecting default assignees by relationship on the case type management screen.

⚠️ **imporant:** this PR makes changes to core by extending its original behaviour. 

## Before
pr-before

## After
pr-after

## Technical Details
A PR fixing this and another issue was sent to core: https://github.com/civicrm/civicrm-core/pull/12647

This PR is being created in order to quickly have the fix in place and not wait for core to approve the PR and a patch PR to be made. 

The fix is done by returning only a list of active relationship types when creating the list of relationship type options. The code was copied from core so it resembles the code in original controller, the only difference is this:

```js
var activeRelationshipTypes = getActiveRelationshipTypes();

function getActiveRelationshipTypes () {
  return apiCalls.relTypes.values.filter(function (relationshipType) {
    return relationshipType.is_active === '1';
  });
}
```

Also, the code was amended in order to work without `underscore`, since it's not available for the extended controller.

Finally, the other fix for missing activity types, as explained by the core PR, can't be fixed using the extended controller since Angular breaks before the extended controller can run.

## Comments

Please note that using labels for comparison of relationship types is not safe, but this PR is following core guidelines:
https://github.com/civicrm/civicrm-core/pull/11998#issuecomment-390243451

The issue is that labels can be the same for multiple relationship types and also can be easily changed by the user. Unfortunately we should not change that core behaviour.

relationship types hr17 9000

-----

✅JS tests ran successfully. Added more tests to cover the new functionality.


uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
```diff
@@ -43,7 +43,50 @@
...
 
       it('initialises the parent controller with activity types having component labels', function () {
-        expect(parentControllerSpy).toHaveBeenCalledWith(expectedApiCalls);
+        expect(parentControllerSpy).toHaveBeenCalledWith(jasmine.objectContaining(expectedApiCalls));
+      });
+
+      describe('default relationship options', function () {
+        var expectedRelationshipOptions;
+
+        beforeEach(function () {
```
igorpavlov: This beforeEach is pretty difficult to follow. Could we send some code out into a function? I think if we move this block https://github.com/compucorp/civihr-tasks-assignments/pull/408/files#diff-cb6f7866f0ad87f9a9f0a4c3669b56c0R60 to a separate func and call it accordingly, that may work.

uk.co.compucorp.civicrm.tasksassignments/js/src/crm-tasks-workflows/controllers/casetype.extended.controller.js
```diff
@@ -53,6 +54,48 @@
...
+     * (Ex: Employee of, Employer is)
+     *
+     * @return {Array}
+     */
+    function getDefaultRelationshipTypeOptions () {
+      var activeRelationshipTypes = getActiveRelationshipTypes();
+      var defaultRelationshipTypeOptions = [];
+
+      activeRelationshipTypes.forEach(function (relationshipType) {
+        var isBidirectionalRelationship = relationshipType.label_a_b === relationshipType.label_b_a;
```
igorpavlov: Are **labels** safe to compare? Just a question, not a CTA.

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
```diff
@@ -157,5 +204,21 @@
         }
       };
     }
+
+    /**
+     * Returns a list of mock relationship types up to the number provided.
+     *
+     * @param {Number} howMany - the number of mock relationship types to return.
```
igorpavlov: Please add an extra space `@param  {Number}` (two spaces between param and Number)

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
```diff
@@ -157,5 +204,21 @@
...
+     * @return {Array}
+     */
+    function getSeveralMockRelationships (howMany) {
```
igorpavlov: Is there any reason we do not have/use relationships mock data?

uk.co.compucorp.civicrm.tasksassignments/js/src/crm-tasks-workflows/controllers/casetype.extended.controller.js
reneolivo: This is the exact same question I made to core. I personally believe it's not safe since they can be changed by the user, but that's what they recommended and what they are using internally. 

Please remember that I'm trying to override a function that actually exists in core, so I'm trying to follow it as closely as possible.

This is my original question and their answer follows soon after:
https://github.com/civicrm/civicrm-core/pull/11998#issuecomment-390243451

At that time, I also double checked and asked through MatterMost.

uk.co.compucorp.civicrm.tasksassignments/js/test/crm-tasks-workflows/controllers/casetype.extended.controller.spec.js
reneolivo: There are several reasons why a mock relationship file was not included:

* There are no mock relationship type files in T&W or Reqangular. In that case one would need to be created, but that's too complex for what this test is trying to accomplish.
* This Angular app purpose is to override core's Angular files. In that sense, it's not connected to Reqangular, T&W, it does not use RequireJS, etc.
* The test does not care much about the relationship type structure, just the `is_active` field. So a function was created to return a simple list of objects with the `is_active` field defined.

uk.co.compucorp.civicrm.tasksassignments/js/src/crm-tasks-workflows/controllers/casetype.extended.controller.js
igorpavlov: Please check if labels can be duped. If not, I would say it is **acceptable** to compare them then.

PR #411, opened 2018-08-20
PCHR-4122: Add Case Type Category Form Field

reneolivo:
## Overview
This PR adds a category field to the Case Type management screen.

## Before
pr-before
(the field does not exists)

## After
pr-after

## Technical Details
First, an upgrader was added to set `Workflow` as the default case type category:

```php
public function upgrade_1042() {
  $caseTypeWorkflowCategory = civicrm_api3('OptionValue', 'get', [
    'sequential' => 1,
    'option_group_id' => 'case_type_category',
    'name' => 'Workflow',
    'option' => ['limit' => 1]
  ]);

  if ($customFields['count'] === 0) {
    return TRUE;
  }

  $caseTypeWorkflowCategory = array_pop($caseTypeWorkflowCategory['values']);

  civicrm_api3('OptionValue', 'create', [
    'id' => $caseTypeWorkflowCategory['id'],
    'is_default' => 1
  ]);

  return TRUE;
}
```

To display the category field, a new CRM's angular hook was added:

```php
$changeSet->alterHtml('~/crmCaseType/caseTypeDetails.html', function (phpQueryObject $doc) {
  _tasksAssignments_add_case_type_category_form_field($doc);
});

function _tasksAssignments_add_case_type_category_form_field (phpQueryObject $doc) {
  $smarty = CRM_Core_Smarty::singleton();
  $formFieldTemplatePath = realpath(__DIR__ . '/views/crm/case-type-category-form-field.tpl');

  $caseTypeCategoryFieldId = CRM_Core_BAO_CustomField::getCustomFieldID(
    'category',
    'case_type_category'
  );
  $caseTypeCategoryOptions = civicrm_api3('OptionValue', 'get', [
    'sequential' => 1,
    'option_group_id' => 'case_type_category',
    'options' => [
      'sort' => 'weight ASC',
      'limit' => 0
    ]
  ]);

  $smarty->assign('caseTypeCategoryFieldId', $caseTypeCategoryFieldId);
  $smarty->assign('caseTypeCategoryOptions', $caseTypeCategoryOptions['values']);

  $formFieldTemplate = $smarty->fetch($formFieldTemplatePath);

  $doc->find('[crm-ui-field*="caseTypeName"]')->after($formFieldTemplate);
}
```

To load the default case type category the route decorator was enhanced:

```js
(function init () {
  var caseTypeEditRoute = $delegate.routes['/caseType/:id'];

[...]

PR #415, opened 2018-08-21
PCHR-4122: Remove category selection from case type management

reneolivo:
## Overview
This PR removes the category form field from the case type management screen and sets "Workflow" as the default category automatically for new case types.

The reason for this is that on this screen we should only create "Workflow" case types and nothing else.

## Before
category-field

## After
pr-after

## Technical Details
Getting the default case type category was removed from the `route.decorator.js` file since it's not going to be used any more. Also, the `case-type-category-form-field.tpl` template file and the code that loaded it were removed because of the same reason.

To select Workflow as the default category for new case types, the `setCaseTypeDefaultCategory` was updated:

```js
function setCaseTypeDefaultCategory () {
  var categoryFieldname = 'custom_' + customFieldIds['caseType.category'];
  var defaultCaseTypeCategory = 'Workflow';

  $scope.caseType[categoryFieldname] = defaultCaseTypeCategory;
}
```

This method is called on initialization and only when creating a new case type. That functionality was introduced in https://github.com/compucorp/civihr-tasks-assignments/pull/411

---

✅Manual Tests - done
✅Jasmine Tests - included and passed
✅JS distributive files - included
⏹CSS distributive files - not needed
⏹Backstop tests - not needed
⏹PHP Unit Tests - not added


uk.co.compucorp.civicrm.tasksassignments/js/src/crm-tasks-workflows/controllers/casetype.extended.controller.js
```diff
@@ -129,11 +129,11 @@
     }
 
     /**
-     * Sets the default category for the case type. The default case type category
-     * is the one selected as default on the API.
+     * Sets Workflow as the default category for the case type.
      */
     function setCaseTypeDefaultCategory () {
       var categoryFieldname = 'custom_' + customFieldIds['caseType.category'];
+      var defaultCaseTypeCategory = 'Workflow';
```
igorpavlov: Should we keep this in a constant? Or at least in the top of the file.

PR #424, opened 2018-09-14
PCHR-3124: Delete Appraisals extension

ajesamson:
## Overview
`uk.co.compucorp.civicrm.appraisals` extension was previously disabled. This PR completely removes the extension.

## Before
Some references to the disabled appraisals extension existed in the Task and Assignment module.

## After
References to the Appraisals in Task and Assignments are now removed.

## Technical Details
Reference to appraisals affects mainly the reminder mechanism in Task and Assignment. Implementations were modified to remove checks for appraisal extension and other related files deleted.


igorpavlov:
I would suggest to name the PR `PCHR-3124: Delete Appraisals extension`

it is more condensed and we all know what it is all about

PR #429, opened 2018-10-03
PCHR-4193: Remove Code Related to Enabling/Disabling Components

tunbola:
## Overview
In https://github.com/compucorp/civihr/pull/2872, the responsibility for enabling and disabling components needed for CiviHR was moved to the HRCore extension. This means that extensions doing this need not do it again. This PR cleans up the code and removes functionality for enabling/disabling the CiviTask and CiviDocument components since the HRCore extension now does this.

## Before
- Code for enabling/disabling the CiviTask and CiviDocument components are present in the extension.

## After
- Code for enabling/disabling the CiviTask and CiviDocument components removed from this extension since the responsibility have been moved to HRCore


uk.co.compucorp.civicrm.tasksassignments/CRM/Tasksassignments/Upgrader.php
```diff
@@ -16,63 +16,12 @@ public function install() {
...
 
   public function enable() {
-    $this->setComponentStatuses(array(
-      'CiviTask' => TRUE,
-      'CiviDocument' => TRUE,
-    ));
-    return TRUE;
   }
 
   public function disable() {
```
davialexandre: Looks like the `enable` and `disable` methods are empty now. Can't we remove them?

